PR #18, opened 2026-02-11
[enh] add flake.nix and nixos, darwin & home module

FlameFlag:
Add Nix Package, NixOS Module, Darwin Module and Home-Manager Module support

I added a GHA Action that's needed for the vendorHash. Anytime deps change a PR will be made with the new hash that just needs to be merged.

Also added `HISTER_DATA_DIR` env var so the modules can set the data dir path depending on the platform (NixOS uses `/var/lib/hister`, Home-Manager uses `~/.local/share/hister`, Darwin uses `~/Library/Application Support/hister`)

I don't know much Go, so sorry if the Go code isn't the optimal way to do it!

P.S: Awesome project, I've been messing around with it since yesterday

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,13 +40,177 @@ Execute `./hister` to see all available commands.
  - Run `./hister help` to list the available commands
  - Execute `./hister listen` to start the web application
 
-
 ## Configuration
 
 Settings can be configured in `~/.config/hister/config.yml` config file - don't forget to restart webapp after updating.
 
 Execute `./hister create-config config.yml` to generate a configuration file with the default configuration values.
 
+
+### Nix
+
+#### Quick usage
+
+Run directly from the repository:
+
+```bash
+nix run github:asciimoo/hister
+```
+
+Add to your current shell session:
+
+```bash
+nix shell github:asciimoo/hister
+```
+
+Install permanently to your user profile:
+
+```bash
+nix profile install github:asciimoo/hister
+```
+
+#### NixOS
+
+Add the following to your `flake.nix`:
+
+```nix
+{
+  inputs.hister.url = "github:asciimoo/hister";
+
+  outputs = { self, nixpkgs, hister, ... }: {
+    nixosConfigurations.yourHostname = nixpkgs.lib.nixosSystem {
+      modules = [
+        ./configuration.nix
+        hister.nixosModules.default
+      ];
+    };
+  };
+}
+```
+
+Then enable the service:
+
+```nix
+services.hister = {
+  enable = true;
+  port = 4433;
+  dataDir = "/var/lib/hister";
+  configPath = /path/to/config.yml; # optional, use existing YAML file
+  config = {  # optional, or use Nix attrset (automatically converted to YAML)
+    app = {
+      directory = "~/.config/hister/";
+      search_url = "https://google.com/search?q={query}";
+    };
+    server = {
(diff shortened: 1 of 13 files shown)


asciimoo:
Thanks!

I'm not really familiar with these bundling solutions except nix; please enlighten me:

Could you explain what is the benefit of having all these in the repo? Why isn't only nix enough?

What maintenance is required to keep them up-to-date?

Is it possible to collect & move all the bundling related scripts/configs/lock files to a dedicated folder under the `assets/` folder to keep the repo root as clean as possible?

hister.go
```diff
@@ -221,6 +221,13 @@ func initialize() {
 
 func initConfig() {
 	var err error
+
+	if cfgFile == "config.yml" {
```
asciimoo: Why do you check if the value of `cfgFile` is `"config.yml"`. If `HISTER_CONFIG` is defined we should use that in every case, shouldn't we?

README.md
```diff
@@ -40,6 +40,170 @@ Execute `./hister` to see all available commands.
  - Run `./hister help` to list the available commands
  - Execute `./hister listen` to start the web application
 
+### Nix
```
asciimoo, line 50: Could you move this section below the configuration section? It is a more commonly needed information in my opinion.

hister.go
4evy: I got a bit confused cuz `cfgFile` defaults to "config.yml" (it's late night for me lol)

will:

`if !rootCmd.PersistentFlags().Changed("config")` 

be okay?

My logic is: 

If the config flag was not changed, then assign cfgFile to envConfig. 

This way `HISTER_CONFIG` is used when user doesn't explicitly pass `--config`, but `--config` flag still has highest priority

asciimoo: > if !rootCmd.PersistentFlags().Changed("config") be okay?

Sure

> If the config flag was not changed, then assign cfgFile to envConfig.

It sounds reasonable, however i would say the same for the opposite as well. =))
I don't know if there is a well defined precedence between ENV vars and command line flags - probably the way you described is the best, having the flag the most powerful.

SeerLite:
Hi, thank you for the making modules @FlameFlag !
I wanted to use them in the meantime but I get type errors.
It looks like the definition for the Home Manager module is broken though, it looks like the systemd services implementation in Home Manager expects a list instead of an attrset like NixOS:
```
error: A definition for option `home-manager.users.seerlite.systemd.user.services.hister.Service.Environment' is not of type `(list of string) or string convertible to it'. TypeError: Definition values:
- In `/nix/store/9f2sr9rh3bq73pr4jid4q3j00x6qg6bv-source/config/nixos/home.nix':
    {
      HISTER_DATA_DIR = "/var/lib/hister";
      HISTER_PORT = "8080";
    }

```

From the man `home-configuration.nix` man page:
```
systemd.user.services.<name>.Service.ExecStart
    Command that is executed when this service is started.

    Type: package or string or list of (package or string)

    Default: [ ]

    Example: "/absolute/path/to/command arg1 arg2"

    Declared by:
        <home-manager/modules/systemd.nix>
```

Also IMO the default value for the data dir shouldn't be /var/lib/hister in the home manager module.

Thank you @asciimoo for hister, it's been great so far!

SeerLite:
@FlameFlag I see, I think you're right then!

PR #19, opened 2026-02-11
[enh] Drop entries from index that are now matching skip rules

clonejo:


server/indexer/indexer.go
```diff
@@ -119,6 +119,11 @@ func Reindex(idxPath, tmpIdxPath string) error {
 				os.RemoveAll(tmpIdxPath)
 				return err
 			}
+			if rules.IsSkip(d.URL) {
+				log.Info().Str("URL", d.URL).Msg("Dropping URL that has since been added to skip rules.")
+				continue
+			}
+			// priority/score are updated implicitly by bleve
```
clonejo, line 126: I am not 100% sure of this, so please double check :)

asciimoo (approved):
Useful addition, thanks!

Issue #3, opened 2026-02-04
Implement result export feature


asciimoo:
Add CSV, JSON, RSS output formats for results.

PR #21, opened 2026-02-12
implemented result export feature --refs issue #3

tfhuhtal:
image

Addded three options for exporting results to csv, json or rss.


asciimoo:
Awesome, thanks!

PR #22, opened 2026-02-12
[enh]: allow multiple URLs in cmd

allendema:
# What
Allow indexing or deleting multiple URLs from cmd at once.

Warn when url is empty.

# How
Loop through the args.


asciimoo:
Useful addition, thanks!

PR #30, opened 2026-02-12
[enh]: support ctrl clicking results

alejoheu:
I noticed that ctrl + clicking a result would not open it in a new tab, which was behaviour I did not expect (most links support this), so I've implemented a workaround for that in this commit. Seems to work on Firefox 147.0.3 at least.


asciimoo:
You are right, thanks!

PR #43, opened 2026-02-14
[mod] refactor nix modules

FlameFlag:
Follow up of https://github.com/asciimoo/hister/pull/18

Addresses: https://github.com/asciimoo/hister/pull/18#issuecomment-3894120666

As well improved the docs (about nix setup) to be more clear, and less verbose

I added `getDefaultDataDir()` to handle platform-specific data directories instead of hardcoding `~/.config/hister` - Also following XDG spec (https://specifications.freedesktop.org/basedir/latest/)

This way it follows `~/Library/Application Support/` on Darwin/MacOS, `XDG_STATE_HOME` on Linux, `LOCALAPPDATA` on Windows, or `UserConfigDir` elsewhere

`StateDirectory` tells `systemd` to create `/var/lib/hister` directory automatically when the unit starts

---

- make `datadir` and `port` optional (default null) to allow config-only
setup
- remove `dataDirMode` option and directory creation logic (it should be
sole responsibility of go)
- move `user`/`group` options to nixos module only
- add shared `histerEnv` helper for environment variables
- update package definition to use flake self reference (so rev is
preserved)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -15,6 +15,7 @@ import (
 	"os/user"
 	"path/filepath"
 	"regexp"
+	"runtime"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -72,6 +73,34 @@ var hotkeyActions = []string{
 	"show_hotkeys",
 }
 
+func getDefaultDataDir() string {
+	switch runtime.GOOS {
+	case "darwin":
+		homeDir, _ := os.UserHomeDir()
+		return filepath.Join(homeDir, "Library/Application Support/hister")
+
+	case "windows":
```
4evy, line 82: `case "windows":`

Technically outside the scope of this PR, but might as well while I'm making this

nix/nixos.nix
```diff
@@ -1,76 +1,60 @@
...
-      };
+      environment = histerEnv config.services.hister;
 
       serviceConfig = {
         ExecStart = "${lib.getExe config.services.hister.package} listen";
         Restart = "on-failure";
         User = config.services.hister.user;
         Group = config.services.hister.group;
-        WorkingDirectory = config.services.hister.dataDir;
+        StateDirectory = "hister";
```
4evy, line 53: This will create `/var/lib/hister`

flake.nix
```diff
@@ -26,26 +26,48 @@
...
       flake = {
         nixosModules.default = inputs.self.nixosModules.hister;
-        nixosModules.hister = import ./nix/nixos.nix;
+        nixosModules.hister =
+          { lib, pkgs, ... }:
+          {
+            imports = [ ./nix/nixos.nix ];
+            services.hister.package = (
+              lib.mkDefault inputs.self.packages.${pkgs.stdenvNoCC.hostPlatform.system}.default
+            );
```
4evy, line 49: The reason the package is being defined from here is so that we can pass it the package from the flake.nix itself, which passes `rev` to the derivation

FlameFlag:
I would appreciate if @SeerLite can review this too

asciimoo:
Great, thanks for both the PR and for the review.

asciimoo:
Is there a better way to get these changes than giving write permission to this github action?

PR #44, opened 2026-02-14
feat(ext): register hister as a search provider

### LINKED ISSUE #37: feature request: have extensions register hister as a search provider
When I added the `hister` extension, I'd like the extension to register `hister` as a search provider. This can be done with the following snippet:
```json
  // in ext/src/manifest.json
  "chrome_settings_overrides": {
    "search_provider": {
      "name": "Hister",
      "keyword": "@hister",
      "search_url": "http://127.0.0.1:4433/?q={searchTerms}",
      "favicon_url": "assets/icons/icon128.png",
      "encoding": "UTF-8",
      "is_default": true 
    }
  },
```
([MDN docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/chrome_settings_overrides#search_provider))

If you're interested, let me know
- what keyword you'd like
- whether the extension should set hister as the default search engine

and I'll drill down on how to dynamically update the search_url based on the backend config and send in a PR. 

(relates to #14)

SKalt:
This PR adds a [`chrome_settings_overrides.search_provider`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/chrome_settings_overrides#search_provider) object to `ext/src/manifest.json`, causing the extension to register the default hister backend as a search provider. I've verified this correctly prompts for consent in Firefox with `is_default: true`: Screenshot from 2026-02-14 14-49-16.
However, I wasn't able to get Chromium to register a new search provider using
```sh
cd ext
npm ci
export PATH="$PWD/node_modules/.bin:$PATH"

npm run build
web-ext run --source-dir ./dist --target chromium
```
Given that installing the Ecosia Chrome extension also failed to register a search provider, this might be due to my laptop's peculiarities. I'll try this code on another machine later.

Additionally, I couldn't find any API that allows dynamically reconfiguring a search provider.

<details open><summary>Things I tried that didn't work</summary>

- the [`search`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/search) api doesn't have any methods for registering search engines
- the [`browser.browserSettings`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserSettings) api lacks support for dynamically registering/updating/deleting search providers, and is only supported by firefox.
- since [`...search_provider.search_url` must be a HTTPS url](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/chrome_settings_overrides#search_url:~:text=This%20must%20be%20an%20HTTPS%20URL), I couldn't use some JS to redirect from an extension-managed homepage

</details>

Since this scratches my immediate itch, this PR closes #37.

### DIFF
--- a/ext/package.json
+++ b/ext/package.json
@@ -16,6 +16,8 @@
         "mini-css-extract-plugin": "^2.9.2",
         "terser-webpack-plugin": "^5.3.11",
         "ts-loader": "^9.5.2",
-        "webpack": "^5.97.1"
+        "web-ext": "^9.3.0",
+        "webpack": "^5.97.1",
+        "webpack-cli": "^6.0.1"
     }
 }
--- a/ext/webpack.config.js
+++ b/ext/webpack.config.js
@@ -90,6 +90,8 @@ const addon = {
                         content = JSON.parse(content.toString());
                         content['version'] = process.env.npm_package_version;
                         content['background']['service_worker'] = 'background.js';
+                        // Chrome throws an error if the favicon URL is not a HTTP URL
+                        contenti['chrome_settings_overrides']['search_provider']['is_default'] = "http://127.0.0.1:4433/static/logo.png"
                         return Buffer.from(JSON.stringify(content));
                     },
                     to: 'manifest.json'
@@ -102,6 +104,7 @@ const addon = {
                         content = JSON.parse(content.toString());
                         content['version'] = process.env.npm_package_version;
                         content['background']['scripts'] = ['background.js'];
+                        content['chrome_settings_overrides']['search_provider']['is_default'] = true
                         content['content_security_policy'] = {
                             "extension_pages": "script-src 'self'"
                         };
--- a/ext/package-lock.json
+++ b/ext/package-lock.json
@@ -1,10 +1,12 @@
 {
-    "name": "Lorch",
+    "name": "Hister",
+    "version": "0.7.0",
     "lockfileVersion": 3,
     "requires": true,
     "packages": {
         "": {
-            "name": "Lorch",
+            "name": "Hister",
+            "version": "0.7.0",
             "devDependencies": {
                 "@types/chrome": "^0.0.299",
                 "clean-webpack-plugin": "^4.0.0",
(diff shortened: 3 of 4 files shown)


ext/webpack.config.js
```diff
@@ -102,6 +102,7 @@ const addon = {
                         content = JSON.parse(content.toString());
                         content['version'] = process.env.npm_package_version;
                         content['background']['scripts'] = ['background.js'];
+                        content['chrome_settings_overrides']['search_provider']['is_default'] = true
```
asciimoo, line 107: Why isn't this `true` by default in the `manifest.json`?

SKalt: Since I couldn't find any documentation I on indicating that chrome would prompt for user consent before setting the default search engine. I'll have a chance to test on a less-weird chrome install later today.

asciimoo: Ah, so the goal is to only set to `true` in the chrome manifest.json. I understand, thanks.

SKalt: Updates from testing with chrome on mac:
- chrome does not prompt for user consent when setting the default search engine using `is_default`
- chrome also requires that the user click to activate the search engine in `chrome://settings/searchEngines`: 
  Screenshot 2026-02-15 at 10 24 20 AM
   This is likely due to the following requirement:

  > Any domain used in the settings API must be [verified](https://support.google.com/webmasters/answer/35179) (via Google Search Console) by the same developer account publishing the extension.
  >
  > -- [https://developer.chrome.com/docs/extensions/reference/manifest/chrome-settings-override](https://developer.chrome.com/docs/extensions/reference/manifest/chrome-settings-override#:~:text=Any%20domain%20used%20in%20the%20settings%20API%20must%20be%20verified%20%28via%20Google%20Search%20Console%29%20by%20the%20same%20developer%20account%20publishing%20the%20extension%2E)
- also, after activating the search engine, chrome forbids altering the search keyword: 
Screenshot 2026-02-15 at 10 24 34 AM (note the grayed-out pencil icon). This suggests there might be some way for the extension to manage the search settings, but I haven't found it yet. 

Chrome's choices about how to do user opt-in kinda cancel out, but I think allowing a user to explicitly opt-in and

ext/package.json
```diff
@@ -16,6 +16,8 @@
         "mini-css-extract-plugin": "^2.9.2",
         "terser-webpack-plugin": "^5.3.11",
         "ts-loader": "^9.5.2",
-        "webpack": "^5.97.1"
+        "web-ext": "^9.3.0",
```
asciimoo, line 19: What is the purpose of this dependency?

SKalt: https://github.com/mozilla/web-ext automates spawning firefox / chrome instances with only the desired extension installed. It's very useful for testing.

asciimoo:
@SKalt unfortunately google rejected the addon with this feature. It says `"The manifest has an invalid search provider url that is not owned by the developer."`. I've temporarily disabled `chrome_settings_override` for chrome. Do you have any idea to make it work with chrome?

SKalt:
It might be possible to have the hister server register dynamically generate the chrome extension manifest and [register itself as an alternate installation/update method](https://developer.chrome.com/docs/extensions/how-to/distribute/host-on-linux#update_url), but that's hacky at best. I think this feature is infeasible on Chrome.

PR #51, opened 2026-02-15
[enh] log which URL is added to index

clonejo:


asciimoo:
Ah, thanks, I did not see fast enough and already added: 798fd1b.
Sorry.

PR #59, opened 2026-02-17
[enh] add sorting by domain

FlameFlag:
image

### DIFF
--- a/server/static/js/search.js
+++ b/server/static/js/search.js
@@ -10,6 +10,11 @@ let urlState = {};
 let lastResults = null;
 let currentSort = "";
 let templates = {};
+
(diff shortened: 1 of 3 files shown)


asciimoo:
Useful feature, but the implementation should be a bit more future proof.

Instead of setting up all the different sort buttons one-by-one, there could be a loop which iterates over the possible sort options (currently it's only `["relevance", "domain"]`). This would make it easier to extend in the future with new sort options (e.g. date or descending order).

If there would be a sort button container (div), a sort button template, a list of possible sort options and a variable storing the currently selected sort option, then the whole sort ui could be automatically generated/updated/extended.

What do you think?

PR #60, opened 2026-02-17
[enh] add navigation, color, and result actions to TUI

FlameFlag:
You can switch between search and results using the `tab` key, navigate with either `j/k` or the arrow keys, and press `enter` to open a result or press `d` to delete it.

Showcase:

https://github.com/user-attachments/assets/b73186ec-3960-4aa9-834d-597f4e63054c


asciimoo (approved):
Great, thanks!

asciimoo:
It looks fancy indeed. I'm not against switching, I didn't know about bubbletea, that's why I choose gocui. =)

The API seems cleaner than gocui as well.

PR #61, opened 2026-02-17
[enh] rewrite TUI using Bubble Tea and add configurable hotkeys

FlameFlag:
Follow up of https://github.com/asciimoo/hister/pull/60#issuecomment-3915375009

Showcase:

https://github.com/user-attachments/assets/f5dc6be3-5b08-4a1a-93bc-2359aa4fe58e

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -495,6 +495,10 @@ func (h Hotkeys) Validate() error {
 }
 
 func (h Hotkeys) ToJSON() template.JS {
+	if h.Web == nil {
+		b, _ := json.Marshal(map[string]string{})
+		return template.JS(b)
+	}
 	b, err := json.Marshal(h.Web)
 	if err != nil {
 		return template.JS("")


asciimoo:
WOOAH! This is epic!

config/config.go
```diff
@@ -49,6 +50,7 @@ type Server struct {
 }
 
 type Hotkeys map[string]string
+type TUIHotkeys map[string]string
```
asciimoo: Wouldn't be better to handle both types of hotkeys under the `Hotkeys` root key, having `Hotkeys.Web` and `Hotkeys.TUI` members?

4evy: Yeah that's a good idea, I'll do it

asciimoo:
You are right, that isn't the change we want, but the hotkey structure change should be adjusted for `search.js`, no?
Probably a `const hotkeys = JSON.parse(document.getElementById("hotkey-data").text).web` is enough.

asciimoo:
@FlameFlag Why did you decided to leave two empty columns on the right side of the ui and one empty row at the bottom?

asciimoo:
huh? for me it is the exact opposite: 
x

Issue #4, opened 2026-02-04
Date Filtering


asciimoo:
Filter results by date

Combine the original query with https://pkg.go.dev/github.com/blevesearch/bleve/v2#NewNumericRangeQuery via https://pkg.go.dev/github.com/blevesearch/bleve/v2#NewConjunctionQuery

PR #64, opened 2026-02-18
[enc] add date filtering

FlameFlag:
Closes: https://github.com/asciimoo/hister/issues/4

### DIFF
--- a/server/indexer/indexer.go
+++ b/server/indexer/indexer.go
@@ -47,6 +47,8 @@ type Query struct {
 	Fields    []string `json:"fields"`
 	Limit     int      `json:"limit"`
 	Sort      string   `json:"sort"`
+	DateFrom  int64    `json:"date_from"`
+	DateTo    int64    `json:"date_to"`
 	base      query.Query
 	boostVal  *query.Boost
 	cfg       *config.Config
@@ -506,6 +508,24 @@ func (q *Query) create() query.Query {
 	}
 
 	q.base = sq
+
+	if q.DateFrom != 0 || q.DateTo != 0 {
+		if q.DateFrom != 0 && q.DateTo == 0 {
+			q.DateTo = time.Now().Unix()
+		}
+		var min, max *float64
+		if q.DateFrom != 0 {
+			min = new(float64)
+			*min = float64(q.DateFrom)
+		}
+		if q.DateTo != 0 {
+			max = new(float64)
+			*max = float64(q.DateTo)
+		}
+		dateQuery := bleve.NewNumericRangeQuery(min, max)
+		dateQuery.SetField("added")
+		q.base = bleve.NewConjunctionQuery(q.base, dateQuery)
+	}
 
 	return q
 }
--- a/server/server.go
+++ b/server/server.go
@@ -288,7 +288,18 @@ func serveIndex(c *webContext) {
 func serveSearch(c *webContext) {
 	q := c.Request.URL.Query().Get("q")
 	if q != "" {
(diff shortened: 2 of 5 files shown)


server/indexer/indexer.go
```diff
@@ -507,6 +509,21 @@ func (q *Query) create() query.Query {
...
+		var min, max *float64
+		if q.DateFrom != 0 {
+			min = new(float64)
+			*min = float64(q.DateFrom)
+		}
+		if q.DateTo != 0 {
+			max = new(float64)
+			*max = float64(q.DateTo)
+		}
+		dateQuery := bleve.NewNumericRangeQuery(min, max)
```
asciimoo, line 525: Why don't you use simply `dateQuery := bleve.NewNumericRangeQuery(float64(q.DateFrom), float64(q.DateTo))`?

server/indexer/indexer.go
```diff
@@ -507,6 +509,21 @@ func (q *Query) create() query.Query {
 
 	q.base = sq
 
+	if q.DateFrom != 0 || q.DateTo != 0 {
```
asciimoo, line 512: If DateFrom > 0 and DateTo == 0, it produces 0 results, no? DateTo should be `now` if it isn't defined.

server/server.go
```diff
@@ -288,7 +288,18 @@ func serveIndex(c *webContext) {
...
+		// Very weird...
+		query := &indexer.Query{Text: q}
+		for param, field := range map[string]*int64{"date_from": &query.DateFrom, "date_to": &query.DateTo} {
+			if v := c.Request.URL.Query().Get(param); v != "" {
+				if t, err := time.Parse("2006-01-02", v); err == nil {
+					*field = t.Unix()
+				}
+			}
+		}
+		r, err := doSearch(query, c.Config)
```
asciimoo, line 302: I think this should be added to the for loop as well.

server/indexer/indexer.go
4evy:
```
$ go doc github.com/blevesearch/bleve/v2 NewNumericRangeQuery

package bleve // import "github.com/blevesearch/bleve/v2"

func NewNumericRangeQuery(min, max *float64) *query.NumericRangeQuery
    NewNumericRangeQuery creates a new Query for ranges of numeric values.
    Either, but not both endpoints can be nil. The minimum value is inclusive.
    The maximum value is exclusive.
```

`NewNumericRangeQuery` takes `*float64` pointers (not values) to support unbounded ranges via nil

The doc says "Either, but not both endpoints can be nil." So I create pointers conditionally to enable this:

```
min := float64(q.DateFrom)
max := float64(q.DateTo)
dateQuery := bleve.NewNumericRangeQuery(&min, &max)
```

Also wouldn't direct conversion `&float64(...)` be invalid?

server/server.go
4evy: `doSearch` is already in the WebSocket loop, since we want one search after parsing both parameters, not multiple searches

server/indexer/indexer.go
4evy: Fixed, added:

```go
if q.DateFrom != 0 && q.DateTo == 0 {
	q.DateTo = time.Now().Unix()
}
```

server/server.go
asciimoo: You are right.

server/indexer/indexer.go
asciimoo: > The doc says "Either, but not both endpoints can be nil."

Alright, mb not reading the docs, sorry.

PR #66, opened 2026-02-18
Allow !! Bang To Work at Start and End

apearson:
## Description 
Currently the !! bang only works as a a prefix `!! hello world`.  This PR allows `hello world !!` to work as well. 

It would make searching with hister and then searching with your external search engine easier as you don't have to move to the start of the search box. 

## Changes
- Updated server.go to check for suffix as well as prefix
- Updated search.js to check fo suffix as well as prefix 

## Testing
- Built hister and tested on Firefox 

Let me know if there any changes you'd like me to make.

- Andrew Pearson

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -264,8 +264,7 @@ func serveIndex(c *webContext) {
 		// Prefix takes precedence over suffix, so "!! query !!" will be treated as "query !!"
 		if strings.HasPrefix(query, "!!") {
 			query = query[2:]
-		}
-		else if strings.HasSuffix(query, "!!") {
+		} else if strings.HasSuffix(query, "!!") {
 			query = query[:len(query)-2]
 		}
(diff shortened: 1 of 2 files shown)


asciimoo:
> It would make searching with hister and then searching with your external search engine easier as you don't have to move to the start of the search box.

You don't have to move to the start either if you use the hotkey to open result with default browser. Why do you prefer this solution over the current hotkey?

apearson:
> > It would make searching with hister and then searching with your external search engine easier as you don't have to move to the start of the search box.
> 
> You don't have to move to the start either if you use the hotkey to open result with default browser. Why do you prefer this solution over the current hotkey?

For me, It's much easier to treat this as a bang similar to DuckDuckGo or Brave search than to learn another shortcut.   I also can't seem to get the hotkeys to work on Firefox on macOS

asciimoo:
I see. Alright, I'm not against this feature.

But the not working hotkeys is a problem. Are you using the latest master?

PR #67, opened 2026-02-18
[fix]: only run workflows in source repo

allendema:
# Why
Only run workflows in source repo and not in every fork, which depending on the workflow creates multiple pull requests.

# How
Check current git repository.

#
Testing welcome.


asciimoo (approved):
Thanks!

PR #73, opened 2026-02-21
Refactor Docker build process and introduce multi-stage distroless images

sinsky:
This PR refactors the Dockerfile to improve security, reduce image size, and provide multiple runtime variants.

Key Changes:
  - Multi-stage Docker Build: Introduced a multi-stage build process using golang:1.24-bookworm as the builder and distroless/base-debian12 as the runtime base. This significantly reduces the final image size and enhances security by removing unnecessary shell utilities and packages.
  - Image Variants: Added support for three distinct Docker targets:
    - release: A secure, non-root variant (default).
    - root: A variant running as the root user for environments requiring elevated permissions.
    - debug: A variant based on the distroless debug image, including a shell for troubleshooting.
  - CI/CD Enhancements: Updated the GitHub Actions workflow (docker-publish.yml) to automatically build and push all three variants with appropriate tags (e.g., :latest, :latest-root, :latest-debug).

Image Size Improvement:
  - Before:(`:latest`) 1.71GB
  - After:(`:latest`) 53.9MB

---

Since the `:latest` tag now uses the non-root variant, this may impact existing users. I considered two options: `:latest` + `:latest-root` vs `:latest` + `:latest-nonroot`, and decided that having non-root as the default would be better for security. Please let me know if you prefer a different approach.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -11,7 +11,17 @@ env:
 
 jobs:
   build-and-push:
+    name: build and push (${{ matrix.target }})
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - target: release
+            suffix: ""
+          - target: root
+            suffix: "-root"
+          - target: debug
+            suffix: "-debug"
     permissions:
       contents: read
       packages: write    # Required to push to GHCR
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Awesome, thanks!

PR #89, opened 2026-02-22
[enh] migrate `search.js` to svelte

FlameFlag:
Move the frontend from vanilla JS to Svelte and TypeScript


asciimoo (approved):
Great, thanks!

PR #91, opened 2026-02-22
Import UX: set --min-visit default to 5, show index stats in footer

doobidoo:
## Summary

- **`--min-visit` default changed from 1 to 5**: Importing with the default of 1 visit results in a large number of low-quality URLs (redirects, one-off visits, error pages) that add noise to the index without providing value. A default of 5 visits better reflects pages the user actually uses.
- **Indexed page count in footer**: The footer now shows the total number of indexed documents (e.g. `1688 pages indexed`), giving users a quick sense of their index size without needing a separate CLI command.
- **App version in footer**: The footer now shows the running version (e.g. `v0.1.0`), useful for support and debugging.

## Changes

- `hister.go`: `--min-visit` flag default `1` → `5`
- `server/indexer/indexer.go`: new `Count() (uint64, error)` function using a size-0 Bleve query
- `server/server.go`: inject `IndexCount` and `AppVersion` into every template render; add `AppVersion` package variable
- `server/templates/layout/base.tpl`: footer extended with `<span class="footer-meta">N pages indexed | vX.Y.Z</span>`

## Test plan

- [ ] Run `hister import chrome <db>` without `--min-visit` flag — should now use threshold of 5
- [ ] Open the web UI — footer should show page count and version
- [ ] Verify `hister import chrome <db> --min-visit 1` still works (explicit override)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/hister.go
+++ b/hister.go
@@ -251,7 +251,7 @@ func init() {
 	listenCmd.Flags().StringP("address", "a", dcfg.Server.Address, "Listen address")
 	indexCmd.Flags().StringP("server-url", "u", dcfg.Server.BaseURL, "hister server URL")
 
-	importCmd.Flags().IntP("min-visit", "m", 1, "only import URLs that were opened at least 'min-visit' times")
+	importCmd.Flags().IntP("min-visit", "m", 5, "only import URLs that were opened at least 'min-visit' times")
 
 	reindexCmd.Flags().BoolP("exclude-sensitive", "x", false, "don't add documents that contain sensitive content matched by config.SensitiveContentPatterns")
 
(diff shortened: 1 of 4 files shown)


asciimoo:
> Importing with the default of 1 visit results in a large number of low-quality URLs (redirects, one-off visits, error pages) that add noise to the index without providing value.

This is subjective. If it does not provide value to you doesn't mean it doesn't provide value for others. You have the ability to change it to 5 if you prefer that, no need to change the default value.

Also, these features are totally unrelated.

Also, we already have a version number.

I ask you again **Please do a thorough review before submitting (especially AI generated) content to respect the time of the maintainers who do code reviews manually.**

I'd also like to ask you to resolve the issue with the already opened PRs before opening new ones.

PR #121, opened 2026-02-25
 feat: migrate to npm workspaces for manging frontend

FlameFlag:


asciimoo (approved):
Fantastic, thanks a lot! As I see most of the features are working as expected. Let's merge it and iterate in smaller steps if we find any broken functionalities.

PR #123, opened 2026-02-27
[Merged] [refactor] tui redesign with tabs, theming, and mouse support

FlameFlag:
The monolithic `tui.go` file was getting massive, so I split it into a much cleaner MVC-like structure under `ui/` (`model`, `handle`, `render`, `network`, and `theme`).

Alongside the refactor, I added a bunch of new features and QoL:

UI and Features:

- Added a multi-tab interface (search, history, rules, add) that you can switch between using alt+1, alt+2, etc
- Full mouse support. you can click tabs, scroll through results, click to focus, and right-click results to open a context menu (open, delete, prioritize)
- Built a base16 theming engine supporting dark/light modes with built-in themes (catppuccin, tokyonight, dracula, nord, gruvbox, rose-pine) and custom yaml loading
- Added interactive overlays for a theme picker (ctrl+t) and keybindings editor (ctrl+s)
- Added the ability to toggle sorting by domain (ctrl+o)
- The rules and add tabs now have fully functional text input forms to manage patterns and manual document indexing directly from the tui

Config and Backend changes:

- Added an optional tui.yml
- Added priority-matching logic so documents matching priority rules are boosted to the top of the search results
- Fixed a bug with QuerySuggestion so it only triggers if the suggestion is actually different from the original query (fixes case-sensitivity issues)
- Updated default keybindings (mapped delete to ctrl+d)

----

- split tui into mvc architecture under ui/
- add multi-tab interface (search, history, rules, add)
- add full mouse support and context menus
- add base16 theming engine and interactive picker
- add keybinding editor overlay
- boost priority-matching documents in search results
- fix query suggestion case-sensitivity bug


server/indexer/indexer.go
```diff
@@ -294,8 +294,10 @@ func (d *Document) Process() error {
 		d.URL = pu.String()
 	}
 	d.Domain = pu.Host
-	if err := d.extractHTML(); err != nil {
-		return err
+	if d.HTML != "" || (d.Text == "" && d.Title == "") {
```
asciimoo: This is not relevant to the topic of the PR, please remove this change. (Also the condition is wrong, it returns true if all `HTML`, `Title` and `Text` members are empty)

server/server.go
```diff
@@ -365,12 +365,28 @@ func doSearch(query *indexer.Query, cfg *config.Config) (*indexer.Results, error
 	if res == nil {
 		res = &indexer.Results{}
 	}
+	// Boost priority-matching documents to the front.
```
asciimoo, line 368: This  is not relevant to the topic of the PR. Also, instead of doing it here it should be done in the indexer during the search. But it pointed out that since the search refactor we do not handle priority  rules at all. Good catch.

ui/handle/actions.go
```diff
@@ -0,0 +1,180 @@
...
+			m.SelectedIdx++
+			render.RefreshViewport(m)
+			m.ScrollToSelected()
+		}
+		return m.FlashHint(config.ActionScrollDown), true
+	case config.ActionDeleteResult:
+		if u := m.GetSelectedURL(); u != "" {
+			m.OpenDeleteDialog("Delete Result", u, -1, func() tea.Cmd {
+				return tea.Batch(
+					network.DeleteURL(m.Cfg.BaseURL(""), u),
```
asciimoo, line 67: Just a note for the future: we should create a client lib that can be used by both the command line tool and the TUI to avoid code duplication.

package-lock.json
```diff
@@ -918,6 +918,7 @@
       "integrity": "sha512-M+MqAvFve12T1HWws/2npP/s3hFtyjw3GB/OXW/8a1jZBk48qnvPJrtgE+VOMc3RnjUMxc4mv/vQ73nvj2uNMg==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
```
asciimoo: This is file should not be part of this PR

asciimoo (commented):
Thank you. I've added a few notes after a first quick check. I'll test it and add further notes if there will be more observations.

ui/handle/actions.go
4evy: True, I could look into that in future PR

config/config.go
```diff
@@ -365,7 +483,63 @@ func (c *Config) Filename() string {
 	return c.FullPath(c.fname)
 }
 
+func (c *Config) SaveConfig() error {
```
asciimoo: This is an unused function, please remove it

server/indexer/indexer.go
```diff
@@ -235,8 +235,24 @@ func Search(cfg *config.Config, q *Query) (*Results, error) {
 		if t, ok := v.Fields["added"].(float64); ok {
 			d.Added = int64(t)
 		}
+		d.Score = v.Score
 		matches[j] = d
 	}
+
+	if q.cfg != nil && q.cfg.Rules != nil {
```
asciimoo: This has nothing to do with the PR, please remove it.

server/indexer/indexer.go
```diff
@@ -294,8 +310,10 @@ func (d *Document) Process() error {
 		d.URL = pu.String()
 	}
 	d.Domain = pu.Host
-	if err := d.extractHTML(); err != nil {
-		return err
+	if d.Text == "" && d.Title == "" {
```
asciimoo: This has nothing to do with the PR, please remove it.

server/server.go
```diff
@@ -370,7 +370,9 @@ func doSearch(query *indexer.Query, cfg *config.Config) (*indexer.Results, error
 		res.History = hr
 	}
 	if oq != "" {
-		res.QuerySuggestion = model.GetQuerySuggestion(oq)
+		if suggestion := model.GetQuerySuggestion(oq); !strings.EqualFold(suggestion, oq) {
```
asciimoo: This has nothing to do with the PR, please remove it.

asciimoo:
Thanks! I've merged it command line.

PR #132, opened 2026-03-01
fix(querybuilder): handle UTF-8 queries in lexer

yshalsager:
## Summary
- switch query lexer input from byte-wise string indexing to rune-aware tokenization
- fix alternation part parsing so the final option is preserved with multibyte text
- add regression tests for Arabic and English tokenization cases

## Validation
- `go test ./server/indexer/querybuilder`
- search manually

image


asciimoo (approved):
Good fix, thank you!

PR #133, opened 2026-03-01
Allow base_url to include subfolder path

### LINKED ISSUE #125: Allow `base_url` config parameter to be a non-root path - e.g. https://x.com/subfolder/
Currently we have hardcoded paths in multiple places so setting up a `base_url` which has a subfolder results in UI errors. Use `base_url` everywhere to resolve relative/absolut paths.

singhAmandeep007:
Fixes #125.

This makes Hister work when server.base_url contains a non-root path prefix (e.g. https://example.com/hister).

Changes:
- Server: accept requests both with and without the base path prefix by stripping it for routing
- Server: treat same-origin correctly for CSRF checks even when base_url includes a path
- Config: include the base path prefix in generated WebSocket URL
- Web UI: inject a runtime <base href> and use base-prefix-safe relative links/assets/API fetches

Tested:
- go generate ./...
- go test ./...

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -35,6 +35,7 @@ type Config struct {
 	SensitiveContentPatterns map[string]string `yaml:"sensitive_content_patterns" mapstructure:"sensitive_content_patterns"`
 	Rules                    *Rules            `yaml:"-" mapstructure:"-"`
 	secretKey                []byte
+	parsedBaseURL             *url.URL
 }
 
 type App struct {
@@ -339,6 +340,10 @@ func (c *Config) init() error {
 		}
 		c.Server.BaseURL = fmt.Sprintf("http://%s", c.Server.Address)
 	}
+	c.Server.BaseURL = strings.TrimSuffix(c.Server.BaseURL, "/")
+	if pu, err := url.Parse(c.Server.BaseURL); err == nil {
+		c.parsedBaseURL = pu
+	}
 	if strings.HasPrefix(c.App.Directory, "~/") {
 		u, _ := user.Current()
 		dir := u.HomeDir
@@ -533,7 +538,7 @@ func (c *Config) BaseURL(u string) string {
 }
 
 func (c *Config) IsSameHost(h string) bool {
-	bu, err := url.Parse(c.BaseURL(""))
+	bu, err := c.baseURLParsed()
 	if err != nil {
 		return false
 	}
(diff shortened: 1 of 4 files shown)


server/server.go
```diff
@@ -211,6 +242,10 @@ func serveSPA(c *webContext) {
 			serve500(c)
 			return
 		}
+		if filepath.Ext(path) == ".html" {
```
asciimoo: This does not feel elegant. Wouldn't be better to pass `base_url` in the `/config` API call?

config/config.go
```diff
@@ -564,10 +565,41 @@ func (c *Config) Host() string {
 }
 
 func (c *Config) WebSocketURL() string {
-	if strings.HasPrefix(c.BaseURL("/"), "https://") {
-		return fmt.Sprintf("wss://%s/search", c.Host())
+	bu, err := url.Parse(c.Server.BaseURL)
```
asciimoo: We could store the parsed `BaseURL` as a private member of the config. This way we could avoid parsing it over-and-over for every request multiple times.

webui/app/src/routes/+layout.svelte
```diff
@@ -31,30 +31,30 @@
...
-    <a data-sveltekit-reload href="/" class="font-outfit text-xl font-extrabold text-text-brand no-underline hover:underline">
+    <img src="static/logo.png" alt="Hister logo" class="h-8 w-8" />
+    <a data-sveltekit-reload href="./" class="font-outfit text-xl font-extrabold text-text-brand no-underline hover:underline">
       Hister
     </a>
   </h1>
   <nav class="flex items-center gap-6">
     <a
-      class:underline={$page.url.pathname === '/history'}
+      class:underline={$page.url.pathname.endsWith('/history')}
```
asciimoo: It would be better to properly check the full path combined with `base_url` to be more future proof and allow other endpoints having these postfixes.

singhAmandeep007: Updated the nav underline logic to compare full resolved paths.

config/config.go
singhAmandeep007: Cached the parsed Server.BaseURL in config to avoid repeated url.Parse calls

server/server.go
singhAmandeep007: Added baseUrl and basePath to /api/config response - so the frontend can discover base URL/path.

PR #137, opened 2026-03-01
[fix] add client module to nix package inputs

antifuchs:
On current main, `nix build github:asciimoo/hister` fails because it can't find the client/ go module. Looks like #94 introduced that module, so it needs to be included in the nix package inputs for the package definition to continue building. This PR adds that input.


asciimoo (approved):
Ah, thanks!

PR #153, opened 2026-03-03
[fix] pass access token from config to HTTP client

### LINKED ISSUE #151: Import don't use --config
During the import, I had to turn off the access token because even with --config <path> it was unable to import anything with 403s.

FlameFlag:
Fixes #151


asciimoo (approved):
Great, thanks!

PR #155, opened 2026-03-03
[enh] add dev server with vite HMR and Go auto-rebuild

FlameFlag:


webui/.air.toml
```diff
@@ -0,0 +1,19 @@
+root = "."
+tmp_dir = "tmp"
+
+[build]
+  entrypoint = "./tmp/hister"
```
asciimoo, line 5: wouldn't be better to use `/tmp/` as the build dir?

4evy: What about Windows? It's also not the standard directory on macOS either

asciimoo: https://pkg.go.dev/os#TempDir

webui/.air.toml
```diff
@@ -0,0 +1,19 @@
+root = "."
+tmp_dir = "$TMPDIR/air"
```
asciimoo: $TMPDIR does not exist on every system

asciimoo: It would be good to default to `/tmp` in this case and even better would be to support defining it from an ENV var.

4evy: I think it would better to go back to `./tmp/hister`, that's how `air` does it anyway

asciimoo: Alright, I don't like the idea of generating junk to the project folder that isn't cleaned up automatically, but it's better than using non-existent folders.

4evy: It gets cleaned up on the exit

```console
> @hister/app@0.5.0 dev
> vite dev

10:01:42 PM [vite] (client) Re-optimizing dependencies because lockfile has changed

  VITE v7.3.1  ready in 1632 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
^Ccleaning...
deleting /Users/flame/Developer/hister/tmp
see you again~
Error: nu::shell::terminated_by_signal

  × External command was terminated by a signal
   ╭─[entry #5:1:1]
 1 │ npm run serve:app
   · ─┬─
   ·  ╰── terminated by SIGSEGV (11)
   ╰────
```

```console
➜ ls ./tmp/
╭───┬────────────┬──────┬──────────┬────────────────╮
│ # │    name    │ type │   size   │    modified    │
├───┼────────────┼──────┼──────────┼────────────────┤
│ 0 │ tmp/hister │ file │ 171.3 MB │ 14 seconds ago │
╰───┴────────────┴──────┴──────────┴────────────────╯
➜ ls ./tmp/
Error: nu::shell::io::not_found

  × Not found
   ╭─[entry #4:1:4]
 1 │ ls ./tmp/
   ·    ───┬──
   ·       ╰── Not found
   ╰────
  help: '/Users/flame/Developer/hister/tmp/' does not exist
```

client/client.go
```diff
@@ -9,10 +9,9 @@ import (
 )
 
 type Client struct {
-	baseURL     string
-	httpClient  *http.Client
-	userAgent   string
-	accessToken string
+	baseURL    string
```
asciimoo: This file is committed by accident IMO

hister.go
```diff
@@ -559,10 +559,6 @@ func importHistory(cmd *cobra.Command, args []string) {
 		if !strings.HasPrefix(u, "http://") && !strings.HasPrefix(u, "https://") {
 			continue
```
asciimoo, line 560: This is a leftover as well. Please check the content of your PR before submitting.

PR #158, opened 2026-03-03
[enh] redesign docs page

FlameFlag:


webui/website/src/content/docs/configuration.md
```diff
@@ -2,6 +2,8 @@
 date: '2026-02-25T00:00:00+00:00'
 draft: false
 title: 'Configuration Reference'
+order: 2
```
asciimoo: It is better to use the date as ordering - or defining a menu manually. Using `order` we have to update a bunch file if we'd like to add something new close to the top.

4evy: The date should just be a representation of when the doc was written or last updated, If we use dates purely for sorting, it will cause a semantic mismatch imo

But thinking about it I guess a better way would be a TypeScript file that maps categories to docs slugs so it's completely decoupled from the content and moving/sorting stuff around is still easy

asciimoo: > But thinking about it I guess a better way would be a TypeScript file that maps categories to docs slugs so it's completely decoupled from the content

This is way better than using `order`

PR #160, opened 2026-03-04
[refactor] extract mouse handling into dedicated package

FlameFlag:


FlameFlag:
The previous code was kind of a mess and very hard to maintain

PR #168, opened 2026-03-06
[enh] add Prettier and EditorConfig across all workspaces

FlameFlag:


package.json
```diff
@@ -3,7 +3,8 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "serve:app": "npm -w @hister/app run serve --"
```
ISSOtm, line 6: Was this removal intended?

4evy: ahh no, my bad

Issue #170, opened 2026-03-06
Horizontal overflow in some docs pages


ISSOtm:
My Firefox viewport is 1295 px wide, and that causes the sidebar to overflow. Actually, the sidebar itself is correctly sized, but the link text overflows.

<img width="1986" height="1233" alt="Image" src="https://github.com/user-attachments/assets/18ec9484-26e5-4a6b-be56-020ec6ba2c66" />

We'd need some `overflow: hidden;` and `text-overflow: ellipsis;` on the sidebar, but I couldn't figure out where it's generated from, so I've been unable to fix it myself.

PR #171, opened 2026-03-06
[fix] wrap long TOC sidebar text instead of overflowing

FlameFlag:
Fixes: https://github.com/asciimoo/hister/issues/170


asciimoo (approved):
Thanks!

PR #182, opened 2026-03-09
[fix] fix all golangci-lint errcheck and unused warnings

FlameFlag:
Basically fix all `golangci-lint` warnings and errors

---

- Fix log.Fatal after defer (exitAfterDefer) defers would never run
- Fix bodyclose: close websocket Dial response body
- Fix file permissions: secret key 0644 -> 0600, rules file ModePerm ->
0600, dirs ModePerm -> 0750

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -56,13 +56,13 @@ func checkStatus(resp *http.Response) error {
 	return fmt.Errorf("invalid status code (%d): %s", resp.StatusCode, msg)
 }
 
-// builds an http.Request with Origin: hister:// set for CSRF bypass.
 func closeBody(resp *http.Response, errp *error) {
 	if cerr := resp.Body.Close(); cerr != nil && *errp == nil {
 		*errp = fmt.Errorf("closing response body: %w", cerr)
 	}
 }
 
+// builds an http.Request with Origin: hister:// set for CSRF bypass.
 func (c *Client) newRequest(method, path string, body io.Reader) (*http.Request, error) {
 	req, err := http.NewRequest(method, c.baseURL+path, body)
 	if err != nil {


client/document.go
```diff
@@ -24,7 +24,7 @@ func (c *Client) AddDocumentJSON(doc *indexer.Document) error {
 	if err != nil {
 		return err
 	}
-	defer resp.Body.Close()
+	defer func() { _ = resp.Body.Close() }()
```
asciimoo: This kind of rewrite does not fix anything, it just makes the code longer. There is still no error handling. The goal of this lint rule is to handle errors everywhere, not to add more boilerplate.

Either ignore this rule or make proper error handling. E.g. you can create a `func closeBody(b *io.Closer)` where you truly check the return value of `resp.body.Close` and log the error.

.golangci.toml
```diff
@@ -10,3 +10,10 @@ module-path = "github.com/asciimoo/hister"
 
 [formatters.settings.goimports]
 local-prefixes = ["github.com/asciimoo"]
+
+[linters.settings.errcheck]
+exclude-functions = [
+  "(net/http.ResponseWriter).Write",
```
asciimoo: It isn't a good idea to disable every network and FS IO checks. The proper way to resolve this issue is to add error handling to the write operations.

ui/model/model.go
```diff
@@ -18,6 +18,8 @@ import (
 	"github.com/asciimoo/hister/server/indexer"
 	"github.com/asciimoo/hister/ui/theme"
 
+	"github.com/rs/zerolog/log"
+
```
asciimoo: no newline required here

client/client.go
```diff
@@ -57,6 +57,12 @@ func checkStatus(resp *http.Response) error {
 }
 
 // builds an http.Request with Origin: hister:// set for CSRF bypass.
```
asciimoo: This comment belongs to the function below. Otherwise LGTM. Please fix this and it's ready to merge

PR #199, opened 2026-03-10
[enh] make ext use same design as app and website

FlameFlag:


webui/ext/src/options/Options.svelte
```diff
@@ -1,6 +1,115 @@
...
+              : 'border-l-hister-rose bg-hister-rose/10 text-hister-rose'}"
+          >
+            {message}
+          </div>
+        {/if}
+
+        <form onsubmit={save} class="space-y-6">
+          <div class="space-y-2">
+            <Label class="font-outfit text-text-brand text-sm font-bold">Server URL</Label>
+            <Input
```
asciimoo: It would be good to create a component/function that produces this input field. The current solution duplicates a lot of code. It makes tedious to change the style.

webui/ext/src/app.css
```diff
@@ -1,2 +1,60 @@
+@import '@fontsource/outfit/400.css';
+@import '@fontsource/outfit/600.css';
+@import '@fontsource/outfit/700.css';
+@import '@fontsource/outfit/900.css';
+@import '@fontsource/fira-code/400.css';
+@import '@fontsource/fira-code/500.css';
+@import '@fontsource/space-grotesk/400.css';
```
asciimoo: Please don't include unused fonts.

webui/ext/src/popup/Popup.svelte
```diff
@@ -89,44 +97,95 @@
...
+    class="border-brutal-border gap-0 rounded-none border-0 border-b-[3px] py-0 shadow-none"
+  >
+    <Card.Header class="bg-muted-surface border-brutal-border border-b-[3px] px-5 py-2.5">
+      <Card.Title class="font-outfit text-text-brand text-xs font-bold tracking-widest uppercase"
+        >Connection</Card.Title
+      >
+    </Card.Header>
+    <Card.Content class="space-y-4 p-5">
+      <form onsubmit={save} class="space-y-4">
+        <div class="space-y-1.5">
```
asciimoo: Same duplicated inputs here. Use `SettingsInput` instead

webui/ext/src/popup/Popup.svelte
```diff
@@ -89,44 +97,88 @@
...
+      <Card.Title class="font-outfit text-text-brand text-xs font-bold tracking-widest uppercase"
+        >Connection</Card.Title
+      >
+    </Card.Header>
+    <Card.Content class="space-y-4 p-5">
+      <form onsubmit={save} class="space-y-4">
+        <SettingsInput
+          label="Server URL"
+          bind:value={url}
+          placeholder="http://127.0.0.1:4433/"
```
asciimoo: I think it is a bit confusing having the default value as a placeholder. It may suggest that the input is already filled. Wouldn't be a better default the original "`Server URL...`" ?

webui/ext/src/popup/Popup.svelte
```diff
@@ -89,44 +97,88 @@
...
+        />
+
+        {#if showTokenInput}
+          <SettingsInput
+            label="Access Token"
+            bind:value={token}
+            placeholder="Optional..."
```
asciimoo: Here the placeholder is misleading as well

asciimoo: Same with the options page

webui/ext/src/popup/Popup.svelte
4evy: Yea, `Server URL...` would be better, I'll go with that

PR #210, opened 2026-03-11
feat: add ability to index and view local files

jamestthompson3:
This PR also adds documentation around running the server on your local machine using either systemd on Linux and launchctl on MacOS


server/indexer/files.go
```diff
@@ -0,0 +1,125 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+	"unicode/utf8"
+
+	"github.com/rs/zerolog/log"
+)
+
+const maxFileSize = 1024 * 1024 // 1MB
```
asciimoo: I'd make this configurable from the config file.

server/indexer/files.go
```diff
@@ -0,0 +1,125 @@
...
+
+func WatchDirectories(dirs []string, interval time.Duration) {
```
asciimoo: Polling can be inefficient with large directories. Wouldn't be better to use https://github.com/fsnotify/fsnotify ?

server/indexer/files.go
```diff
@@ -0,0 +1,125 @@
...
+func indexAll(dirs []string) {
+	for _, dir := range dirs {
+		dir = expandHome(dir)
+		if err := indexDirectory(dir); err != nil {
+			log.Error().Err(err).Str("directory", dir).Msg("Failed to index directory")
+		}
+	}
+}
+
+func expandHome(path string) string {
```
asciimoo: This is a general purpose function that we need in `config.go` and that you use in `server.go` as well. Perhaps could be moved to a package to make it accessible from both places.

server/indexer/files.go
```diff
@@ -0,0 +1,125 @@
...
+		return fmt.Errorf("cannot access directory: %w", err)
+	}
+	if !info.IsDir() {
+		return fmt.Errorf("not a directory: %s", dir)
+	}
+
+	indexed := 0
+	skipped := 0
+
+	log.Info().Str("directory", dir).Msg("Indexing directory")
```
asciimoo: This is a debug level message to me.

server/server.go
```diff
@@ -536,6 +537,15 @@ func doSearch(query *indexer.Query, cfg *config.Config) (*indexer.Results, error
 	if oq != "" {
 		res.QuerySuggestion = model.GetQuerySuggestion(oq)
 	}
+	if len(cfg.Indexer.Directories) > 0 {
+		basePrefix := cfg.BasePathPrefix()
+		for _, doc := range res.Documents {
+			if strings.HasPrefix(doc.URL, "file://") {
+				filePath := strings.TrimPrefix(doc.URL, "file://")
+				doc.URL = basePrefix + "/api/file?path=" + url.QueryEscape(filePath)
```
asciimoo: You can use `cfg.BaseURL("/api/...")` here.

webui/website/src/content/docs/service-management.md
```diff
@@ -0,0 +1,161 @@
...
+
+After building or installing a new version of hister, restart the service to pick up the changes:
+
+```bash
+launchctl kickstart -k gui/$(id -u)/org.hister.search
+```
+
+---
+
+## Linux (systemd)
```
asciimoo: I'd change the order of the mac/linux section. I think, there are more linux/systemd Hister setups.

webui/website/src/content/docs/service-management.md
```diff
@@ -0,0 +1,161 @@
+---
```
asciimoo: This file have to be added to `webui/website/src/lib/docs-structure.ts`.

jamestthompson3:
I like this idea. I can work on it.

webui/website/src/content/docs/service-management.md
```diff
@@ -0,0 +1,162 @@
...
+systemctl --user enable hister
+
+# Start the service now
+systemctl --user start hister
+```
+
+To have user services start at boot (even before logging in):
+
+```bash
+sudo loginctl enable-linger $USER
```
ISSOtm: This is a bad idea: this enables far more than hosted itself, and can cause breakage. 

We will ship a system unit ourselves (see https://codeberg.org/asciimoo/hister/pulls/12 for my WIP; main open question for me: which directory to place them in the repo? Some generic `etc`, or a dedicated `systemd`?), and I think this user version should be shipped inline also, so as to prevent it from drifting with changes to hister itself. (Note that this would ideally require some CI, but I'm not sure how to test that. Try starting the server via systemd and check that the client can communicate with it?)

We can then document how to make config override via a drop-in file (https://wiki.archlinux.org/title/Systemd#Editing_provided_units), though a user unit will just use the default config file anyway so that may be more appropriate.

webui/website/src/content/docs/service-management.md
```diff
@@ -0,0 +1,162 @@
...
+title: 'Running as a Service'
+---
+
+To keep Hister running in the background and start it automatically on boot, you can set it up as a system service. Below are instructions for Linux (systemd) and macOS (launchctl).
+
+## Linux (systemd)
+
+### 1. Create the service file
+
+Create `~/.config/systemd/user/hister.service`:
```
ISSOtm: It may be preferable to use `systemd --user edit --full --force hister.service`, as that automatically performs the necessary daemon reload. (Though this is generally moot if we ship the user unit ourselves.)

webui/website/src/content/docs/service-management.md
```diff
@@ -0,0 +1,162 @@
...
+
+```ini
+[Unit]
+Description=Hister Search Engine
+After=network.target
+
+[Service]
+Type=simple
```
ISSOtm: Why not the `exec` type?

webui/website/src/content/docs/service-management.md
```diff
@@ -0,0 +1,162 @@
...
+ExecStart=/path/to/hister listen
+Restart=on-failure
+RestartSec=5
```
ISSOtm: Why five seconds?

webui/website/src/lib/docs-structure.ts
```diff
@@ -22,7 +22,7 @@ export const docsStructure: DocsCategory[] = [
   },
   {
     name: 'Advanced Server Setup',
-    slugs: ['server-setup', 'docker'],
+    slugs: ['server-setup', 'docker', 'service-management'],
```
ISSOtm: In my draft PR, I've made a page specifically for systemd and another one for launchctl, since they share no information. What do you think?

webui/website/src/content/docs/service-management.md
```diff
@@ -0,0 +1,162 @@
...
+systemctl --user start hister
+```
+
+To have user services start at boot (even before logging in):
+
+```bash
+sudo loginctl enable-linger $USER
+```
+
+### 3. Managing the service
```
ISSOtm: I feel like we may be served better by deferring to an external resource, like the Arch wiki, since that is much more comprehensive. What do you think?

PR #218, opened 2026-03-14
Harden file watcher and indexer with proper skip lists, symlink checks, and cleaner server helpers

FlameFlag:
The file watcher was essentially broken, the debounce logic never actually
fired callbacks because it was iterating timers in a separate goroutine that
would block forever on the first one. This rewrites `WatchDirectories` to use
`time.AfterFunc` with a proper mutex so debounced writes actually get processed

It also takes a context now so the caller can shut it down cleanly, and returns
an error instead of silently swallowing startup failures

While in there, the watcher now skips hidden directories and a comprehensive
list of well-known build/dependency directories (`node_modules`, `__pycache__`,
`target`, `vendor`, etc.) so we're not burning inotify watches on stuff nobody
wants indexed. The exclude patterns from the config are also applied during the
directory walk now, not just at file-match time

On the server side, `serveFile` now resolves symlinks before checking whether a
path falls within a configured directory, which closes a path traversal vector
where a symlink inside an allowed directory could point anywhere on disk

The CSRF origin check in `withCSRF` got flattened from nested ifs into a single
boolean expression, and the duplicated MIME detection logic got pulled into
a small `detectMIME` helper. The bang-bang redirect also got simplified using
CutPrefix/CutSuffix instead of the HasPrefix-then-TrimPrefix dance

The file indexer now uses sentinel errors (`ErrEmptyFile`, `ErrBinaryFile`,
`ErrFileTooLarge`) and an atomic for `maxFileSize` so it's safe to call from
multiple goroutines

The rest is formatter and linter cleanup

### DIFF
--- a/files/files.go
+++ b/files/files.go
@@ -87,12 +87,14 @@ var skipDirs = map[string]struct{}{
 // shouldSkipDir reports whether a directory should be excluded from watching.
 // It skips hidden directories, well-known dependency/cache directories, and
 // directories matching any exclude pattern from the config.
-func shouldSkipDir(name string, excludes []string) bool {
-	if strings.HasPrefix(name, ".") {
-		return true
-	}
-	if _, ok := skipDirs[name]; ok {
-		return true
+func shouldSkipDir(name string, excludes []string, includeHidden bool) bool {
+	if !includeHidden {
+		if strings.HasPrefix(name, ".") {
+			return true
+		}
+		if _, ok := skipDirs[name]; ok {
+			return true
+		}
 	}
 	for _, pattern := range excludes {
 		if matched, _ := filepath.Match(pattern, name); matched {
@@ -100,6 +102,11 @@ func shouldSkipDir(name string, excludes []string) bool {
 		}
 	}
 	return false
+}
+
+// ShouldSkipDir is the exported form of shouldSkipDir for use by the indexer.
+func ShouldSkipDir(name string, excludes []string, includeHidden bool) bool {
+	return shouldSkipDir(name, excludes, includeHidden)
 }
 
 // walkAndWatch registers all subdirectories of each configured directory with
@@ -119,7 +126,7 @@ func walkAndWatch(watcher *fsnotify.Watcher, dirs []config.Directory) {
 			if !d.IsDir() {
 				return nil
 			}
-			if path != expanded && shouldSkipDir(d.Name(), excludes) {
+			if path != expanded && shouldSkipDir(d.Name(), excludes, dir.IncludeHidden) {
 				return filepath.SkipDir
 			}
 			if err := watcher.Add(path); err != nil {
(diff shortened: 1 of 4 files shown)


files/files.go
```diff
@@ -68,71 +71,202 @@ func findMatchingDir(dirs []config.Directory, filePath string) *config.Directory
 	return nil
 }
 
-func WatchDirectories(dirs []config.Directory, callback func(string)) {
+// skipDirs lists non-hidden directory names that should never be watched.
+// Hidden directories (starting with ".") are already skipped unconditionally,
+// which covers .git, .svn, .hg, .bzr, .venv, .tox, .cache, .gradle, .idea,
+// .vscode, .next, .nuxt, .terraform, .mypy_cache, .pytest_cache, .ruff_cache,
+// .bundle, .stack-work, .cabal-sandbox, .dart_tool, .pub-cache, .ccache, etc.
+var skipDirs = map[string]struct{}{
```
asciimoo, line 79: There is no need for this, please remove this list. Users have options to exclude files/diretories. Don't restrict them - especially don't do it with generic folder names like `eggs`, `wheels`, `out`, and most importantly `logs`.

server/indexer/files.go
```diff
@@ -1,19 +1,32 @@
...
-	"github.com/rs/zerolog/log"
 )
 
-var maxFileSize int64 = 1024 * 1024 // 1MB default
+var (
+	ErrEmptyFile    = errors.New("empty file")
+	ErrBinaryFile   = errors.New("binary file")
+	ErrFileTooLarge = errors.New("file too large")
+
+	maxFileSize atomic.Int64
```
asciimoo: No need for atomic here. We write this only once at startup.

files/files.go
```diff
@@ -68,71 +71,202 @@ func findMatchingDir(dirs []config.Directory, filePath string) *config.Directory
...
+	}
+	for _, pattern := range excludes {
+		if matched, _ := filepath.Match(pattern, name); matched {
+			return true
+		}
+	}
+	return false
+}
+
+func WatchDirectories(ctx context.Context, dirs []config.Directory, callback func(string)) error {
```
asciimoo: This function became too large. Please refactor it to smaller components.

files/files.go
4evy: Yeah, stuff like `eggs`, `wheels`, `out`, `logs` is too generic, I'll remove them

Although I still want to keep the list for sane defaults like `node_modules`, trying to index will blow through the `inotify` watch limit very fast

What about dot folders (and their files)? Most of time it's stuff we _don't_ want to index and it's meant to be "hidden" by default anyways?

files/files.go
4evy: Makes sense. I'll pull out the initial directory walk into its own helper and extract the per-event handling so the main loop is just a dispatcher

FlameFlag:
> The rest is formatter and linter cleanup... Please remove off-topic changes

Sure, but please run `npm run format` later

> This isn't relevant here.

Fair

files/files.go
asciimoo: > Although I still want to keep the list for sane defaults like node_modules, trying to index will blow through the inotify watch limit very fast

You are right, this is a sensible default. In the mean time I'd let users overwrite it with an `include_all` / `include_hidden_files` boolean flag in the directory config.

> What about dot folders (and their files)? Most of time it's stuff we don't want to index and it's meant to be "hidden" by default anyways?

Same thing. It's good to exclude by default but allow users to include it, no?

asciimoo:
> Sure, but please run npm run format and golangci-lint  later tho

Good point, I've fixed it.

Issue #157, opened 2026-03-03
Feature: Add Sync and Queue


FlameFlag:
At the moment, if the server is down and you visit a page, it won't be sent when the server comes back online this is pretty major, as for example, if the server silently goes down you won't know, and there will be no way to get that history back unless you're willing to re-import all your history again, but this could upset other stuff you don't want. 

It will also be nice to have a feature where, when the server goes down or comes back online, you get notified from the extension

simcop2387:
This would also be really nice for when you're like me, running hister on the server in your closet but don't have it exposed to the internet at large so when you go somewhere it's not accessible you don't lose any history.

PR #223, opened 2026-03-15
[feat] add offline queue and batch API for browser extension

FlameFlag:
Closes: https://github.com/asciimoo/hister/issues/157

Before this, if the Hister server was unreachable when you visited a page, that
page was just silently dropped, you'd never know it wasn't indexed.

This PR fixes that by giving the extension a proper offline queue backed by
IndexedDB (via Dexie)

When a send fails, the page or history item gets stored locally. A Chrome alarm
fires every two minutes and tries to "drain" the queue whenever it sees the
server is back online. You can also hit "Retry now" manually from the popup

Items that keep failing stay in the queue until you clear or remove them
individually

To make draining efficient, the server now has a `/api/batch` endpoint that
accepts up to 100 documents and history items in a single request. The CORS
bypass that already existed for `/api/add` was extended to cover `/api/batch`
as well

The search page also picks up queued items now. When you search, the content
script asks the background worker to run the query against the local queue and
sends any matches back to the page, where they show up with a "Pending sync"
badge so you know they haven't hit the server yet

On the extension side there's a new `QueuePanel` component for the popup that
shows connection status, how many pages are waiting, and lets you expand the
list to see or remove individual items

### DIFF
--- a/webui/app/src/routes/+page.svelte
+++ b/webui/app/src/routes/+page.svelte
@@ -17,7 +17,7 @@
     openURL,
   } from '$lib/search';
   import { fetchConfig, apiFetch } from '$lib/api';
-  import type { SearchResults } from '$lib/search';
+  import type { SearchResults, SearchResult } from '$lib/search';
   import { animate } from 'animejs';
   import { Input } from '@hister/components/ui/input';
   import { Button } from '@hister/components/ui/button';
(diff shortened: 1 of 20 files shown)


webui/app/src/routes/+page.svelte
```diff
@@ -203,6 +203,23 @@
     autocomplete = (query && res.query_suggestion) || '';
     highlightIdx = 0;
     resultsShown = true;
+
+    // Request supplemental results from extension's offline queue
+    if (query) {
+      window.dispatchEvent(new CustomEvent('hister:search', { detail: { query } }));
```
asciimoo, line 209: This is a big security risk. Any website can get access to the results this way.

PR #226, opened 2026-03-15
[fix] nix/package.nix add `files` package to build inputs

FlameFlag:


asciimoo:
Do we have to manually update the hash each time we add a new package to the code?

PR #233, opened 2026-03-19
[feat] add POST `/api/batch` endpoint for bulk add/delete/get operations

FlameFlag:
For now, there's no direct usage; however, this will be useful for the extension in the future

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -933,6 +933,7 @@ func serveBatch(c *webContext) {
 		return
 	}
 
+	batch := indexer.NewMultiBatch()
 	results := make([]batchOpResult, len(req.Ops))
 	for i, op := range req.Ops {
 		switch op.Op {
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -884,6 +884,98 @@ func serveDeleteDocument(c *webContext) {
...
+
+	results := make([]batchOpResult, len(req.Ops))
+	for i, op := range req.Ops {
+		switch op.Op {
+		case "add":
+			if op.URL == "" {
+				results[i] = batchOpResult{Status: http.StatusBadRequest, Error: "missing url"}
+				continue
+			}
+			d := &indexer.Document{URL: op.URL, Title: op.Title, Text: op.Text}
```
asciimoo: We need the the other fields as well. Favicon; HTML for reindexing purposes.

server/server.go
```diff
@@ -884,6 +884,98 @@ func serveDeleteDocument(c *webContext) {
...
+}
+
+type batchResponse struct {
+	Results []batchOpResult `json:"results,omitempty"`
+	Error   string          `json:"error,omitempty"`
+}
+
+const maxBatchOps = 100
+
+func serveBatch(c *webContext) {
```
asciimoo, line 918: The whole point of batch processing is to use Bleve batch processing support. This function executes the operations one-by-one instead of using our `multiBatch`

server/server.go
```diff
@@ -884,6 +884,98 @@ func serveDeleteDocument(c *webContext) {
 	serve200(c)
 }
 
+type batchOp struct {
+	Op    string `json:"op"`
```
asciimoo: The possible values for the `Op` could be constants.

PR #241, opened 2026-03-19
[fix] prevent Svelte 5 proxy from being stored in Chrome storage for `customHeaders`

FlameFlag:
Use `$state.snapshot()` when saving `customHeaders` to `chrome.storage.local` and add `Array.isArray` guards on load to handle any previously-corrupted values


asciimoo (approved):
Thanks!

PR #245, opened 2026-03-20
[feat] add generic enricher interface, Properties field, and Media DocType

FlameFlag:
Some websites just don't work well when you store them as plain HTML.

YouTube is the most obvious one, the actual value of having a YouTube video in your history, is knowing what the video was about, who made it, how long it was, and ideally what was said in it. Same thing applies to say TikTok, Twitch clips, SoundCloud, etc.

This adds an **opt-in** "enrichment" (disabled by default, needs `enrichment.enabled: true` in config) that uses `yt-dlp` to pull metadata for URLs matching "known" media domains, things like title, channel, duration, description, view/like counts, categories, chapters, and optionally subtitles/transcripts. The enriched data replaces the raw HTML in the index, which means these entries become actually searchable by their real content. You can find a video by something that was said in it, by the channel name, by topic, whatever.

It runs in the background via a worker pool so it doesn't block normal indexing. The metadata gets stored as JSON in the existing `html` field and survives re-indexing without needing to hit `yt-dlp` again.

---

I understand there are privacy implications with making external requests to fetch metadata. That said, this is and should always stay an opt-in feature, it's off by default and you have to explicitly enable it. 

Realistically, if you're already visiting these sites in your browser, making an additional request to grab metadata isn't meaningfully different from a privacy standpoint. I do intend to add proxy support down the line as well. For now though, this is still pretty early and experimental.

One way to tackle the privacy aspect that we could look into is to query Invidious instances instead of YouTube, but this is more of an idea than something I plan on doing.

### DIFF
--- a/server/enricher/manager.go
+++ b/server/enricher/manager.go
@@ -76,10 +76,20 @@ func (m *Manager) Enqueue(url, domain string) {
 	}
 }
 
-// Close signals workers to stop and waits for them to finish.
+// Close signals workers to stop, waits for in-flight work to finish,
+// and drains any remaining items in the queue.
 func (m *Manager) Close() {
 	close(m.done)
 	m.wg.Wait()
+	// Drain any items that were queued but not yet picked up by a worker.
+	for {
+		select {
+		case item := <-m.queue:
+			m.process(item)
+		default:
+			return
+		}
+	}
 }
 
 func (m *Manager) worker() {


asciimoo:
> I kept it as a separate interface from the existing Extractor rather than extending it directly, the reasoning being that extractors are fundamentally synchronous (they run inline during Document.Process() on the HTML that's already there), while enrichers are async by nature since they need to call out to external tools in the background. 

I think it's better to view them as both are responsible to extract content from the sources. It just happens that this particular extractor that you try to implement is preferred to work async. This does not make it that much different in my opinion.
If there is an extractor interface and all the extractors are dedicated sub packages, then it is up to the extractors how they want to operate, it won't cause any issues outside their package as long as they implement the `Extractor` interface.

I imagine the following structure:
```
extractor/ <- extractor package that implements the `Extractor` interface and imports the`extractors`
extractor/extractors/ <- collector folder for all the different extractors
extractor/extractors/readability/ <- an existing extractor implementation
extractor/extractors/ytdlp <- same as the previous
...
```

> Mixing the two execution models into one interface felt like it would add unnecessary complexity to the existing extractors

I see it differently.

- Both features are triggered when new data comes in
- Both parse the data
- Both create `Document`s
- Both have to support `Document` related operations

There are much more similarities than differences.

Providing multiple types of solution to one problem is what makes the software unnecessarily complex. A well thought-out interface that is flexible to support multiple use-cases can even simplify the software's architecture.

> The worker pool manager is still there (it made sense as-is), but now it's generic

This is a great approach. We could provide reusable helper building blocks that can be imported and used by multiple extractors. If an extractor wants to operate asynchronously, it can use the `WorkerPoolManager` and so on.

server/extractor/extractor.go
```diff
@@ -0,0 +1,96 @@
...
+	FaviconURL string         `json:"favicon_url,omitempty"`
+	Properties map[string]any `json:"properties,omitempty"`  // display metadata (key-value pairs)
+	StoredData string         `json:"stored_data,omitempty"` // JSON blob stored in HTML field (survives reindex)
+}
+
+// Extractor defines the interface for content extraction.
+// Implementations may operate synchronously (e.g., HTML parsing) or
+// asynchronously (e.g., fetching metadata from external tools).
+// Async extractors can use the Manager as a reusable worker pool helper.
+type Extractor interface {
```
asciimoo, line 18: The extractor interface does not feel very versatile.

I'd add an `Initialize(config) error` function to be able to fine tune each extractor from the config. 

Why did you decide to restrict the `Match` and `Extract` functions to use only limited amount of data instead of keeping `Document` as their argument? This change introduces unnecessary new types and it limits the available resources an extractor can use. What if I'd like to write a language specific extractor? Or if I'd like to process only public (uid 0) results?

The `Rebuild` method is currently unused and not entirely clear to me what is the purpose. It would be better to add it only after an extractor requires it. Probably it will provide better explanation about its purpose as well.

config/config.go
```diff
@@ -79,6 +80,12 @@ type Indexer struct {
 	MaxFileSize     int64        `yaml:"max_file_size_mb" mapstructure:"max_file_size_mb"`
 }
 
+type Enrichment struct {
```
asciimoo: You mean Extractor. Also, perhaps it would be better to add it to the indexer instead of to the root, no?

server/extractor/extractor.go
```diff
@@ -0,0 +1,96 @@
...
+
+// ErrRebuildNotSupported is returned by extractors that do not support
+// rebuilding from stored data.
+var ErrRebuildNotSupported = errors.New("rebuild not supported")
+
+const extractorKey = "_extractor"
+
+// IsStoredExtraction checks whether the HTML field contains extraction
+// data rather than regular HTML content.
+func IsStoredExtraction(html string) bool {
```
asciimoo: The current general unwritten policy is that we keep the HTML untouched it is just stored to always have the original source material and modify the `Text` attribute by the extractors. Keeping the HTML immutable allows us to reindex content even if we change/reconfigure extractors.

server/extractor/manager.go
```diff
@@ -0,0 +1,130 @@
+package extractor
```
asciimoo: Why do we need a central manager? Why can't each extractor handle their own processing how they want? Most of the extractors don't even need a manager and a queue handler, but others maybe prefer different ways/queue sizes/parallel processes than others. For me it would make more sense to provide building blocks that can any extractor can import and use if they want and how they want.

Do not create abstractions prematurely, while there isn't at least two common things that can be hid behind an abstraction. Otherwise it just becomes unnecessary boilerplate.

server/indexer/document.go
```diff
@@ -19,22 +20,48 @@ import (
...
+	Title              string         `json:"title"`
+	Text               string         `json:"text"`
+	Favicon            string         `json:"favicon"`
+	Score              float64        `json:"score"`
+	Added              int64          `json:"added"`
+	Type               types.DocType  `json:"type"`
+	Language           string         `json:"language"`
+	UserID             uint           `json:"user_id"`
+	Properties         map[string]any `json:"properties,omitempty"`
+	PropertiesRaw      string         `json:"properties_raw,omitempty"`
```
asciimoo: Why do we need raw properties? Can't bleve store objects?

server/server.go
```diff
@@ -908,7 +965,7 @@ func serveReadable(c *webContext) {
 	if r.Title() != "" {
 		title = r.Title()
 	}
-	c.JSON(map[string]string{
+	c.JSON(map[string]any{
```
asciimoo: Why do you change it to `any`?

server/server.go
```diff
@@ -158,6 +160,43 @@ func Listen(cfg *config.Config) {
 		HttpOnly: true,
 	}
 
+	if cfg.Enrichment.Enabled {
+		extractorManager = extractor.NewManager(
+			cfg.Enrichment.Workers,
+			cfg.Enrichment.Timeout,
+			func(url string, result *extractor.Result, existing *extractor.ExistingDoc) error {
+				d := &indexer.Document{
```
asciimoo: The issue with recreating the Document from a custom result type this way is that we need to update it every time we update the Document type. This already lacks the `UserID` field.

server/extractor/extractor.go
asciimoo: how does this comment got resolved? It still does not use the Document type but creates an "Input" type which mimics the properties of the "Document" type. These duplicated types does not help maintainability.

server/enricher/manager.go
```diff
@@ -0,0 +1,129 @@
...
+	return m.enrichers
+}
+
+// Enqueue checks if any registered enricher matches the URL and, if so,
+// adds it to the processing queue.
+func (m *Manager) Enqueue(url, domain string) {
+	for _, e := range m.enrichers {
+		if e.Match(url, domain) {
+			select {
+			case m.queue <- queueItem{url: url, domain: domain}:
```
asciimoo, line 70: What happens if the queue gets full or if the server shuts down before the queue got empty?

PR #253, opened 2026-03-23
Add User handling/multiuser support

asciimoo:
This PR introduces user management to Hister.

Users can be deleted/created via the command line `create-user` and `delete-user` commands.
This feature is fully backward compatible and turned off by default. To enable user management, add `user_handling: true` to the configs `app` section.

### Extension changes

To simplify the extension authentication handling, the addon fetches the Hister webapp cookies and uses those in network requests instead of prompting the user for a token and/or username/password. 

Is there any security consequences of this decision?

changes:
- Removed access token authentication from the extension entirely
- chrome.cookies.getAll() used to fetch session cookies for the configured server URL and inject them as a Cookie header on API requests
- Authenticate Extension button added in popup and options page: fetches and caches Hister webapp cookies in chrome.storage.local, then verifies against /api/profile; button is hidden when already authenticated and appears on 403 or after failed auth check
- "cookies" permission added to manifest

### New endpoints

- `POST /api/login` authenticates with username/password, sets session cookie
- `POST /api/logout`  clears session
- `GET /api/profile` returns user info when authenticated; returns empty 200 for token/no-auth mode; 403 when unauthenticated
- `POST /api/profile/token` generates a new per-user access token (CSRF protected)
- `GET /api/config`  now returns authMode (token | user | none) and username

@FlameFlag @scumjr I'd appreciate a thorough review, because this is a substantial change that have security impacts.

UPDATE:

Currently the password handling does not use salt. We could use the value of the `Config.SecretKey()` which reads the content of the `[config_dir]/.secret_key` file. Should we use it as a salt? The reason against it is if the file is accidentally deleted, then the user authentication becomes completely inaccessible.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -34,13 +34,14 @@ import (
 )
 
 var (
-	appSubFS         iofs.FS
-	staticFileServer http.Handler
-	sessionStore     *sessions.CookieStore
-	errCSRFMismatch  = errors.New("CSRF token mismatch")
-	storeName        = "hister"
-	tokName          = "csrf_token"
-	staticTextFiles  map[string][]byte
+	appSubFS                 iofs.FS
+	staticFileServer         http.Handler
+	sessionStore             *sessions.CookieStore
+	errCSRFMismatch          = errors.New("CSRF token mismatch")
+	storeName                = "hister"
+	tokName                  = "csrf_token"
+	staticTextFiles          map[string][]byte
+	multiUserNotSupportedMsg = map[string]string{"error": "rule management is not yet supported in multi-user mode"}
 )
 
 type historyItem struct {
@@ -844,8 +845,7 @@ func serveRules(c *webContext) {
 		return
 	}
 	if c.Config.App.UserHandling {
-		// TODO
-		serve500(c)
+		c.JSONStatus(http.StatusNotImplemented, multiUserNotSupportedMsg)
 		return
 	}
 	err := c.Request.ParseForm()
@@ -1038,7 +1038,7 @@ func serveOpensearch(c *webContext) {
 func serveAddAlias(c *webContext) {
 	if c.Config.App.UserHandling {
 		// TODO
-		serve500(c)
+		c.JSONStatus(http.StatusNotImplemented, multiUserNotSupportedMsg)
 		return
 	}
 	err := c.Request.ParseForm()
@@ -1062,7 +1062,7 @@ func serveAddAlias(c *webContext) {
 func serveDeleteAlias(c *webContext) {
 	if c.Config.App.UserHandling {
 		// TODO
-		serve500(c)
+		c.JSONStatus(http.StatusNotImplemented, multiUserNotSupportedMsg)
 		return
 	}
 	err := c.Request.ParseForm()


server/model/user.go
```diff
@@ -0,0 +1,76 @@
...
+	Password string `json:"-"`
+	Token    string `json:"-"`
+}
+
+func CreateUser(username, password string) (*User, error) {
+	var existing User
+	if err := DB.Where("username = ?", username).First(&existing).Error; err == nil {
+		return nil, ErrUserAlreadyExists
+	}
+	hash, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
```
4evy, line 33: bcrypt already generates and stores a per-hash random salt, so you don't need to add one manually

If you want an extra layer (a "pepper"), that's a separate decision, but the current code is sound from a password hashing standpoint imo

webui/ext/src/modules/network.ts
```diff
@@ -14,20 +14,32 @@ async function fetchFavicon(url) {
   });
 }
 
-async function sendPageData(url, doc, tok, customHeaders = []) {
+async function getServerCookies(): Promise<string> {
```
4evy, line 17: The session cookies are grabbed via `getAll()` and stored as a raw string. While `chrome.storage.local` is sandboxed per-extension (so other extensions can't read it), it might be worth only grabbing the session cookie by name rather than all cookies, to avoid storing unrelated cookies from the same domain

server/server.go
```diff
@@ -425,6 +467,69 @@ func serveSPA(c *webContext) {
...
+	delete(session.Values, "user_id")
+	delete(session.Values, "username")
+	if err := session.Save(c.Request, c.Response); err != nil {
+		serve500(c)
+		return
+	}
+	serve200(c)
+}
+
+func serveProfile(c *webContext) {
```
4evy, line 540: I traced the middleware ordering here - `populateUserContext` runs first in `createHandler` (fills `c.UserID` from the session/token), then `withUserAuth` checks `c.UserID == 0` and returns 403 if unauthenticated. Since the Profile endpoint doesn't set `NoAuth: true`, unauthenticated requests should correctly get blocked before reaching `serveProfile`.

The flow looks correct to me, but could you confirm with a quick curl that an unauthenticated request to `/api/profile` actually gets a 403 when `user_handling` is enabled? Just want to double check.

server/server.go
```diff
@@ -425,6 +467,69 @@ func serveSPA(c *webContext) {
 	serveIndex(c)
 }
 
+func serveLogin(c *webContext) {
```
4evy, line 497: There's no rate limiting on `/api/login`. A simple counter (e.g., sleep after N failed attempts per IP or per username) would go a long way. Not necessarily a blocker, but worth considering imo

server/model/user.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return nil, err
+	}
+	u := &User{Username: username, Password: string(hash), Token: rand.Text()}
+	return u, DB.Create(u).Error
+}
+
+func DeleteUser(username string) error {
```
4evy, line 41: When a user is deleted, their indexed documents and history entries stick around with the orphaned `user_id`. Is that intentional? 

Should there be an option to purge a user's data on deletion, or at least reassign it to a "global" user (user_id=0)?

server/api.go
```diff
@@ -213,5 +215,38 @@ func init() {
 			Handler:      serveAPI,
 			Description:  "API documentation",
 		},
+		{
+			Name:         "Login",
+			Path:         "/api/login",
+			Method:       POST,
+			CSRFRequired: false,
```
4evy: Login CSRF is a minor concern here, with `CSRFRequired: false` and `NoAuth: true`, an attacker could potentially force a victim to log into the attacker's account? (This is more of a theoretical) 

Maybe worth setting a CSRF token on the login page load as a follow-up?

server/indexer/indexer.go
```diff
@@ -577,6 +583,17 @@ func (q *Query) create() query.Query {
...
 	}
 
+	if q.UserID > 0 {
+		uid := float64(q.UserID)
+		userQuery := bleve.NewNumericRangeInclusiveQuery(&uid, &uid, boolPtr(true), boolPtr(true))
+		userQuery.SetField("user_id")
+		zeroF := float64(0)
+		globalQuery := bleve.NewNumericRangeInclusiveQuery(&zeroF, &zeroF, boolPtr(true), boolPtr(true))
+		globalQuery.SetField("user_id")
+		userOrGlobal := bleve.NewDisjunctionQuery(userQuery, globalQuery)
```
4evy, line 615: This treats `user_id=0` as "global" so all authenticated users can see those documents. Makes sense for backward compatibility (existing pre-multiuser docs will have `user_id=0`), but worth adding a comment explaining the semantics

Also, if `user_handling` is enabled but someone somehow hits the API without auth, their docs get `user_id=0` and become visible to everyone

webui/ext/src/background/background.ts
```diff
@@ -16,10 +16,9 @@ function clearErrorBadge(tabId: number) {
 // TODO check source
 function cjsMsgHandler(request, sender, sendResponse) {
   chrome.storage.local
-    .get(['histerURL', 'histerToken', 'indexingEnabled', 'histerCustomHeaders'])
+    .get(['histerURL', 'indexingEnabled', 'histerCustomHeaders'])
```
4evy: The extension no longer sends `X-Access-Token` directly - it now relies on cookies. For token-auth mode (no `user_handling`), this means users have to first log in via the web UI to get a session cookie, then use "Authenticate Extension" to grab it. This still works, but it's a workflow change that should probably be documented. Could the extension keep the token input as a fallback for token-only setups?

hister.go
```diff
@@ -284,6 +286,60 @@ var deleteCmd = &cobra.Command{
...
+		}
+		initDB()
+	},
+	Run: func(cmd *cobra.Command, args []string) {
+		username := args[0]
+		password, err := promptPassword("Password: ")
+		if err != nil {
+			exit(1, "Failed to read password: "+err.Error())
+		}
+		if password == "" {
```
4evy: The only validation here is that the password isn't empty. Might be worth enforcing a minimum length (e.g., 8 chars)?

server/indexer/indexer.go
asciimoo: > Also, if user_handling is enabled but someone somehow hits the API without auth

I'm not sure how can this happen. If `user_handling` is enabled then you cannot add pages to the index without authentication.

server/model/user.go
asciimoo: It isn't intentional, however I'm not sure what is the best way to handle it. Neither of the suggested resolutions are good defaults. Deletion can lead to data loss and making it global can expose sensitive data.

Should we add a `--purge`  and `--assign=<new_user_id>` flags to the `delete` command and let the admin decide it?

server/server.go
asciimoo: Double check done it responds with 403

webui/ext/src/modules/network.ts
asciimoo: > it might be worth only grabbing the session cookie by name rather than all cookie

This way we'd need to update the extension every time we introduce a new cookie that is useful/required to send it from the extension. The differences are marginal (especially because we don't have other cookies currently), so I keep it this way to be more flexible and we can reiterate it in the future.

server/server.go
asciimoo: This is a good idea on the long run, but I would not include it to this PR. There are good solutions to handle these kind of issues app-independently ( https://github.com/fail2ban/fail2ban )

server/server.go
```diff
@@ -425,6 +493,72 @@ func serveSPA(c *webContext) {
...
+		return
+	}
+	session, err := sessionStore.Get(c.Request, storeName)
+	if err != nil {
+		serve500(c)
+		return
+	}
+	session.Values["user_id"] = user.ID
+	session.Values["username"] = user.Username
+	session.Values["is_admin"] = user.IsAdmin
```
4evy: `is_admin` is stored in the session at login time and `populateUserContext` reads it straight from the session without hitting the DB again. if an admin gets demoted via `update-user --toggle-admin`, their existing sessions still carry `is_admin: true` until they log out (which could be up to a year given session lifetime).

worth either re-validating admin status from the DB on each request (or with a short TTL), or invalidating sessions when admin status changes. this is probably the most security-relevant thing in the PR imo

server/server.go
```diff
@@ -698,6 +845,11 @@ func serveRules(c *webContext) {
 		serve500(c)
 		return
 	}
+	if c.Config.App.UserHandling {
+		// TODO
```
4evy: the TODO stubs returning 500 here (and in `serveAddAlias`/`serveDeleteAlias`) will look like a server error to users with no indication of what went wrong. something like a 501 or a JSON body `{"error": "rule management is not yet supported in multi-user mode"}` would be a lot friendlier

server/model/user.go
```diff
@@ -0,0 +1,120 @@
...
+	Token    string `json:"-"`
+	IsAdmin  bool   `json:"is_admin"`
+}
+
+func CreateUser(username, password string, isAdmin bool) (*User, error) {
+	var existing User
+	if err := DB.Where("username = ?", username).First(&existing).Error; err == nil {
+		return nil, ErrUserAlreadyExists
+	}
+	hash, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
```
4evy, line 33: minor: bcrypt silently truncates passwords longer than 72 bytes, so a 200 character password would hash identically to its first 72 bytes. not critical but maybe worth either documenting or adding a max length check with a clear error

server/server.go
```diff
@@ -927,7 +1095,8 @@ func serveDeleteDocument(c *webContext) {
 	if fp, err := url.QueryUnescape(strings.TrimPrefix(u, c.Config.BaseURL("/api/file?path="))); err == nil && fp != u {
 		u = "file://" + fp
 	}
-	if err := indexer.Delete(u); err != nil {
+	docID := indexer.GetDocID(c.UserID, u)
```
4evy, line 1096: this scopes deletion to the current user's docs which is correct, but it also means admins can't delete another user's documents through the API. might be fine for v1 but worth noting as a potential follow-up

server/indexer/indexer.go
```diff
@@ -602,6 +604,17 @@ func (q *Query) create() query.Query {
...
```
4evy, line 615: a short comment here explaining the `user_id=0` semantics would help future readers. something like "user_id=0 docs are global/pre-multiuser and visible to all authenticated users for backward compat"

webui/ext/src/popup/Popup.svelte
```diff
@@ -34,17 +32,49 @@
...
+      authURL += '/';
+    }
+    const doCheck = (cookies: string) => {
+      const headers: HeadersInit = { 'Content-Type': 'application/json' };
+      if (cookies) {
+        headers['Cookie'] = cookies;
+      }
+      return fetch(authURL + 'api/profile', { headers })
+        .then((r) => {
+          isAuthenticated = r.status === 200;
```
4evy: if the server is unreachable (network issue, server down), this sets `isAuthenticated = false` and shows the "Authenticate Extension" button, which is a bit misleading since the user might already have valid credentials. could be worth distinguishing "not authenticated" from "can't reach server" so users don't keep re-authenticating for no reason

server/model/user.go
4evy: yeah i think that's the right call, letting the admin decide explicitly is better than picking a default that could silently lose data or leak it.

one thought: maybe the command should refuse to delete if the user has any documents/history and no flag is passed? that way nobody accidentally orphans data without realizing. something like "user has N documents, use --purge or --assign to handle them"

server/indexer/indexer.go
4evy: yeah fair enough, i traced through the middleware again and `withUserAuth` blocks the add endpoint so there's no way to hit it unauthenticated. my bad on that second part.

the comment suggestion still stands though, the `user_id=0` as "global" behavior is non-obvious enough that a one-liner in the query code would save someone a head scratch later

server/server.go
asciimoo: The whole admin delete functionality requires an overhaul and I think it should be handled independently in a custom function.
Admins should have options to specify which users they want to delete from or if they want to delete an URL from every user. I've added this to my TODO list, but I don't want to delay further this PR by implementing here. Do you agree?

4evy: > Do you agree?

Sure

PR #267, opened 2026-03-31
Add missing Args to API endpoints (Login, Add, GenerateToken)

rsharma2be23-cloud:
### Summary

This PR adds missing `Args` definitions to several API endpoints in `server/api.go` to improve API consistency and documentation.

### Changes

* Added required arguments for:

  * Login endpoint (`username`, `password`)
  * Add document endpoints (`url`, optional `title`)
  * GenerateToken endpoint (`description`)
* Ensured consistency across alias endpoints (`/api/add` and `/add`)

### Impact

* Improves API clarity and self-documentation
* Enables better validation and future extensibility

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -83,6 +83,20 @@ func init() {
 			CSRFRequired: true,
 			Handler:      serveAdd,
 			Description:  "Save added document",
+			Args: []*EndpointArg{
+				{
+					Name:        "url",
+					Type:        "string",
+					Required:    true,
+					Description: "URL of the document to add",
+				},
+				{
+					Name:        "title",
+					Type:        "string",
+					Required:    false,
+					Description: "Optional title of the document",
+				},
+			},
 		},
 		// alias for /api/add - backward compatibility - use /api/add in the future
 		{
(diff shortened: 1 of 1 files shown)


server/api.go
```diff
@@ -92,6 +106,20 @@ func init() {
 			CSRFRequired: true,
 			Handler:      serveAdd,
 			Description:  "Save added document",
+			Args: []*EndpointArg{
```
asciimoo, line 109: The function accepts a `Document` object or `url`, `title` and `text` fields. Currently there is no easy way to define the `Document` object, but `text` is still missing.

Also, these are the same arguments as the previous function's arguments. Should be defined in one place and referenced in both endpoints to avoid code duplication.

server/api.go
```diff
@@ -264,6 +306,14 @@ func init() {
 			CSRFRequired: true,
 			Handler:      serveGenerateToken,
 			Description:  "Generate a new access token for the current user",
+			Args: []*EndpointArg{
```
asciimoo, line 309: There is no such argument for this endpoint.

asciimoo (changes requested):
There are many endpoints still missing and the endpoints added have inaccuracies.

PR #307, opened 2026-04-09
[enh] add concurrency controls, graceful shutdown, and thread safety …

boromil:
…to file indexing

- Add worker pool with bounded channel for file indexing and watching
- Add context cancellation and graceful shutdown to IndexAll and WatchDirectories
- Add signal handling (SIGINT/SIGTERM) with clean HTTP server shutdown
- Add configurable max_watch_dirs and index_workers settings
- Add SyncSet and SyncMap concurrency primitives in xsync package
- Add thread-safe language indexer creation with LoadOrCompute
- Add watch directory limit to prevent file descriptor exhaustion
- Add batched index writes with periodic flushing during initial indexing
- Fix missing IndexAlias registration when creating language-specific indexes
- Fix thread-safety of indexer map access with sync.Map-backed SyncMap


asciimoo:
@boromil?

PR #319, opened 2026-04-13
[feat] add yt-dlp extractor

FlameFlag:
Add a new yt-dlp extractor that fetches video metadata (title, description, chapters, thumbnails, subtitles) for YouTube, Vimeo, TikTok, and other supported platforms

The extractor is **disabled by default** and requires `yt-dlp` to be installed

Enable it in config (e.g):

```yaml
extractors:
  ytdlp:
    enable: true
    options:
      binary: yt-dlp               # path to binary
      timeout: 15                  # seconds
      fetch_subtitles: false       # download & index transcripts
      sub_language: en
      cookies_file: ""             # path to cookies.txt
      cookies_from_browser: ""     # e.g. "chrome"
      extra_args: []               # additional CLI flags
```

### DIFF
--- a/webui/app/src/routes/+page.svelte
+++ b/webui/app/src/routes/+page.svelte
@@ -28,6 +28,7 @@
   import * as DropdownMenu from '@hister/components/ui/dropdown-menu';
   import * as Tooltip from '@hister/components/ui/tooltip';
   import { ScrollArea } from '@hister/components/ui/scroll-area';
+  import VideoPreview from '$lib/components/VideoPreview.svelte';
   import { Kbd } from '@hister/components/ui/kbd';
   import {
     Search,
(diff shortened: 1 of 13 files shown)


webui/app/src/routes/+page.svelte
```diff
@@ -81,15 +82,28 @@
   let showPopup = $state(false);
   let popupTitle = $state('');
   let popupContent = $state('');
+  let popupTemplate = $state('');
+  let popupVideoData = $state<any>(null);
```
asciimoo: It would be good to use a more generic way to handle extractor data. E.g. `popupTemplateData` variable instead of a video specific one.

webui/app/src/routes/+page.svelte
```diff
@@ -81,15 +82,28 @@
...
   let actionsQuery = $state('');
   let actionsMessage: string | null = $state(null);
   let actionsError = $state(false);
   let showActionsForResult: string | null = $state(null);
 
+  function parseVideoData(content: string): any | null {
```
asciimoo: Same as `popupVideoData`. Be generic.

asciimoo (approved):
Awesome, thanks!

PR #328, opened 2026-04-15
[fix] make file indexing work on Windows

### LINKED ISSUE #327: Windows - File Indexing not working
Hello =), 

It seems that file-indexing is not working on Windows.
I added a "directory" and defined the "file-types" in the config files. 
Stopped Hister and restarted it.

Config was like this:

path was 'c:\user\xy\'
filetypes was ['txt']
detect languages: false

Image

But if I search for type:file or type:loca or for the filename itself, noting shows up.

cheers

FlameFlag:
Closes #327

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -887,7 +887,9 @@ func (d *Directory) IsMatching(name string) bool {
 	}
 	if len(d.Filetypes) > 0 {
 		ext := strings.TrimPrefix(filepath.Ext(name), ".")
-		if !slices.Contains(d.Filetypes, ext) {
+		if !slices.ContainsFunc(d.Filetypes, func(ft string) bool {
+			return strings.EqualFold(ft, ext)
+		}) {
 			return false
 		}
 	}
--- a/files/files.go
+++ b/files/files.go
@@ -34,11 +34,44 @@ func ExpandHome(path string) string {
 // Debounce so we don't spam the index as write events can file multiple times before closing a file after editing
 const debounceTime = 200 * time.Millisecond
 
+// HasPathPrefix reports whether filePath equals dirPath or is contained within it,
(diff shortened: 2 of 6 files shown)


asciimoo:
What was the root cause of the bug and how does this PR fix it?

FlameFlag:
The root cause is that the indexing code was written assuming Unix paths

Directory containment checks were hardcoded to `/`, e.g. `strings.HasPrefix(filePath, dirPath+"/")`. A file at `C:\Users\xy\notes.txt` never matches the configured dir `C:\Users\xy` because the separator is `\`. Fixed by a `HasPathPrefix` helper using `filepath.Separator`, reused in `FindMatchingDir` and `serveFile`.

Second, `file://` URLs were built with plain concatenation: `"file://" + absPath`. On Windows that produces `file://C:\Users\xy\notes.txt`, which isn't a valid file URL. The reverse conversions were equally broken, one spot stripped `file://` with `CutPrefix`, another used `url.Parse(...).Path`, and both mangled the drive letter since the input was malformed to begin with.

Fixed it by routing every conversion through two helpers, `PathToFileURL` (`C:\foo` -> `file:///C:/foo`) and `FileURLToPath` (the inverse)

Third, the filetype check was case-sensitive: `slices.Contains(d.Filetypes, ext)`. Windows filesystems aren't, so `Notes.TXT` wouldn't match a configured `txt`. Swapped for `slices.ContainsFunc` with `strings.EqualFold`

After these fixes the exact config from the issue (`path: c:\user\xy\`, `filetypes: [txt]`) indexes correctly

asciimoo:
Oh, good to know. Thanks for the explanation!

asciimoo (approved):
Thanks!

PR #329, opened 2026-04-15
[fix] use `sendBeacon` for history on result click

### LINKED ISSUE #177: Fix middle click and ctrl+click behavior by rewriting opened results save (saveHistory) to use navigator.sendBeacon
Currently the result opening is very messy. We prevent the default click event to be able to save the query and the opened result in the backend (`openResult` & `saveHistory` functions). This causes middle click/ctrl+click to work differently. The new tab opened immediately gets the focus instead of opening in the background just like with standard middle/ctrl+click.

After many experiments and stackoverflow questions it seems like it isn't possible to open a new tab in the background from js.

The solution to this issue could be not preventing the default link click event and  using https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon to send back the url/title/query in the background.

FlameFlag:
Closes #177

Stop preventing the default link click so middle/ctrl/cmd/shift+click behave natively (new tab opens in the background, shift opens a new window, etc)

History is saved via navigator.sendBeacon with an application/json Blob, which fires reliably even as the page navigates away on a plain left click. The existing backend handler is unchanged; it decodes the JSON body regardless of how it was delivered

### DIFF
--- a/webui/app/src/routes/+page.svelte
+++ b/webui/app/src/routes/+page.svelte
@@ -237,6 +237,16 @@
     saveHistoryItem(url, stripHtml(title), query, false, () => openURL(url, newWindow));
   }
 
+  function sendHistoryBeacon(url: string, title: string, queryStr: string) {
+    const payload = JSON.stringify({
+      url,
+      title: stripHtml(title),
+      query: queryStr,
+      delete: false,
+    });
+    navigator.sendBeacon('api/history', new Blob([payload], { type: 'application/json' }));
+  }
+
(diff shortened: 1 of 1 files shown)


asciimoo:
Are you sure that middle/ctrl+click is triggered by `onClick` this way?

asciimoo (approved):
Nice, thanks!

asciimoo:
Please format the code and it's ready to merge.

asciimoo (approved):

PR #330, opened 2026-04-15
[fix] tighten `aws_access_key` sensitive pattern

### LINKED ISSUE #318: Sensitive data regex is overzealous
e.g `aws_access_key: AKIA[0-9A-Z]{16}` matching base64, e.g CSS fonts in https://www.koalick.info/immobilien-gmbh/aktuelle-projekte, which combined with #317 creates user confusion (is hister server not responding?)

FlameFlag:
Closes #318

The AKIA[0-9A-Z]{16} regex matched inside base64 blobs and CSS font data, producing false positives

Require whitespace, quote, or string boundary around the token

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -356,7 +356,7 @@ func CreateDefaultConfig() *Config {
 			},
 		},
 		SensitiveContentPatterns: map[string]string{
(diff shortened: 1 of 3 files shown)


asciimoo (approved):
Nice, thanks!
I appreciate the extra tests <3

PR #331, opened 2026-04-15
[fix] dedupe priority results vs standard search results

### LINKED ISSUE #2: Merge standard and priority results with the same URLs
Priority results (green results) can be shown as duplicates of normal results. Merge matching normal results into the priority result and remove the normal result from the results lists to avoid duplicated display of the same URL.

FlameFlag:
Closes #2

When a URL appears in both the priority (history) list and the standard search results, drop the standard entry so it isn't shown twice. The priority entry keeps its position and styling

### DIFF
--- a/server/model/history.go
+++ b/server/model/history.go
@@ -35,6 +35,7 @@ type HistoryLink struct {
 type URLCount struct {
 	URL   string `json:"url"`
 	Title string `json:"title"`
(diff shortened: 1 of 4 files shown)


asciimoo (changes requested):
It would be good to add the normal result's text to the priority result, because priority results don't have body by default.

asciimoo (approved):
Nice, thanks!

PR #332, opened 2026-04-15
[enh] add opensearch suggestions endpoint

### LINKED ISSUE #14: Suggestions URL for Firefox
In Firefox, it's possible to add something called suggestion URL for search engines,
so the engine can populate the suggestion field in the searchbar.

Does Hister have anything like this yet?

Image

FlameFlag:
Closes #14

Expose /suggest returning the OpenSearch suggestions JSON format (`[query, [title, ...]]`) so browsers like Firefox can fetch live completions from their search bar. The opensearch.xml descriptor advertises the new URL alongside the existing HTML search template.

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -67,6 +67,14 @@ func init() {
 			Handler:      serveSearch,
 			Description:  "Search websocket endpoint",
 		},
+		{
+			Name:         "Suggest",
+			Path:         "/suggest",
+			Method:       GET,
+			CSRFRequired: false,
+			Handler:      serveSuggest,
+			Description:  "OpenSearch suggestions endpoint",
+		},
 		// tmp added for backward compatibility
 		{
 			Name:         "Add",
--- a/server/server.go
+++ b/server/server.go
@@ -1099,6 +1099,15 @@ func serveOpensearch(c *webContext) {
 const suggestLimit = 10
 
 func serveSuggest(c *webContext) {
+	// Sec-Fetch-Site is set by browsers and forbidden to JS, so a cross-site
+	// fetch() can't spoof it. Browser address-bar flows either omit the header
+	// (Firefox) or send "none" (Chrome); reject anything explicitly cross-site.
+	switch c.Request.Header.Get("Sec-Fetch-Site") {
+	case "", "none", "same-origin", "same-site":
+	default:
+		c.Response.WriteHeader(http.StatusForbidden)
+		return
+	}
 	q := c.Request.URL.Query().Get("q")
 	suggestions := []string{}
 	if q != "" {
--- a/server/server.go
+++ b/server/server.go
@@ -213,6 +213,7 @@ func registerEndpoints(cfg *config.Config) http.Handler {
 	mux.HandleFunc("GET /static/", createHandler(cfg, serveStatic))
 	mux.HandleFunc("GET /favicon.ico", createHandler(cfg, serveFavicon))
 	mux.HandleFunc("GET /opensearch.xml", createHandler(cfg, serveOpensearch))
+	mux.HandleFunc("GET /suggest", createHandler(cfg, serveSuggest))
 	mux.HandleFunc("/", createHandler(cfg, serveSPA))
 	// If base_url contains a non-root path prefix (e.g. https://x.com/subfolder),
 	// accept requests both with and without that prefix.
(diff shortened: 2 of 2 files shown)


server/server.go
```diff
@@ -213,6 +213,7 @@ func registerEndpoints(cfg *config.Config) http.Handler {
 	mux.HandleFunc("GET /static/", createHandler(cfg, serveStatic))
 	mux.HandleFunc("GET /favicon.ico", createHandler(cfg, serveFavicon))
 	mux.HandleFunc("GET /opensearch.xml", createHandler(cfg, serveOpensearch))
+	mux.HandleFunc("GET /suggest", createHandler(cfg, serveSuggest))
```
utafrali: `/suggest` is registered directly with `mux.HandleFunc` outside the `Endpoints` loop (lines 195-210), which is where auth middleware (`withTokenAuth`, `withUserAuth`) is applied. This means the endpoint is publicly accessible even when the instance has `AccessToken` configured or `UserHandling` enabled.

Worse, when an unauthenticated request reaches `serveSuggest`, `c.UserID` stays at `0`. Looking at `indexer/indexer.go:733`, the user filter is only added when `q.UserID > 0` -- so a request with `UserID: 0` runs against the full index across all users, leaking every user's browsing history as autocomplete suggestions.

The fix is to drop the manual `mux.HandleFunc` call and register this in `api.go` alongside the other endpoints:

```go
{
    Name:         "Suggest",
    Path:         "/suggest",
    Method:       GET,
    CSRFRequired: false,
    Handler:      serveSuggest,
    Description:  "OpenSearch suggestions endpoint",
},
```

The existing loop will then apply the right auth wrapper automatically.

server/server.go
```diff
@@ -1088,13 +1089,44 @@ func serveOpensearch(c *webContext) {
...
+			log.Warn().Err(err).Msg("suggest search failed")
+		}
+		if res != nil {
+			for _, d := range res.Documents {
+				title := strings.TrimSpace(d.Title)
+				if title == "" {
+					title = d.URL
+				}
+				suggestions = append(suggestions, title)
+			}
```
utafrali, line 1129: `json.NewEncoder(c.Response).Encode(...)` writes a trailing newline after the JSON. For the two-field OpenSearch format this is harmless in practice, but all other JSON responses in this file use `json.Marshal` + `c.Response.Write(jr)`, which does not append a newline. Keeping the pattern consistent avoids a subtle discrepancy:

```go
jr, err := json.Marshal([]any{q, suggestions})
if err != nil {
    log.Warn().Err(err).Msg("failed to marshal suggest response")
    return
}
if _, err := c.Response.Write(jr); err != nil {
    log.Warn().Err(err).Msg("failed to write suggest response")
}
```

utafrali (commented):
The feature itself is well-scoped and the opensearch.xml change is correct, but `/suggest` is wired outside the auth middleware loop, making it publicly accessible on any token-protected or multi-user instance and leaking the full search index to unauthenticated callers. This needs to be moved into the `Endpoints` slice before merging.

server/server.go
asciimoo: @utafrali thanks for the review, I agree.

FlameFlag:
Thanks a lot for the review!

server/api.go
```diff
@@ -67,6 +67,14 @@ func init() {
 			Handler:      serveSearch,
 			Description:  "Search websocket endpoint",
 		},
+		{
+			Name:         "Suggest",
+			Path:         "/suggest",
+			Method:       GET,
+			CSRFRequired: false,
```
asciimoo, line 74: I wonder how big security risk it is to bypass the CSRF verification. In theory, this way an attacker could potentially steal private URLs.
But, the issue with turning on the CSRF validation is that it won't work from the browser's URL bar, so it defeats the purpose of this feature.
I'm not sure yet how to solve this situation.

4evy: `Sec-Fetch-Site` might be a workable gate here, per the [Fetch Metadata spec and MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Fetch-Site), browser-issued requests (address bar, bookmarks, system-principal fetches) send `Sec-Fetch-Site: none, and Sec-*` headers are forbidden to JS so an attacker page can't spoof them

Firefox specifically fixed OpenSearch requests to use none in bug [1703466](https://bugzilla.mozilla.org/show_bug.cgi?id=1703466)

So instead of CSRF, we could gate /suggest on:

```go
site := r.Header.Get("Sec-Fetch-Site")
if site != "" && site != "none" && site != "same-origin" && site != "same-site" {
    http.Error(w, "forbidden", http.StatusForbidden
     return
}
```

That keeps the URL-bar flow working while blocking cross-site reads.

One caveat is that...: the Mozilla bug explicitly covers the OpenSearch description fetch; suggestion requests behave the same by analogy (same system-principal path), but I haven't found an authoritative doc confirming it for Chrome/Safari. 

Probably worth logging the header on /suggest for a day and hitting it from the address bar in each browser before enforcing. If any browser sends cross-site, we'd need a different approach (e.g., a per-user opaque token baked into the OpenSearch descriptor URL)

asciimoo: > Sec-Fetch-Site might be a workable gate here

The question is that the browser sends this header when calling suggest from the URL bar or not.

> Probably worth logging the header on /suggest 

Yes this feature requires thorough testing on the major browsers.

> per-user opaque token baked into the OpenSearch descriptor URL

It would just push the problem to another endpoint. This case the opensearch descriptor would require CSRF to protect this token from external access, which leads to the same issue.

4evy: Okay so I ran a small header-logging server locally and exercised `/suggest` from the address bar in both browsers

Firefox 149 sends no `Sec-Fetch-*` headers at all on the suggest fetch (the navigation that follows when you press Enter does send Sec-Fetch-Site: none, but the suggestion XHR is on a different code path)

Chrome 147 sends Sec-Fetch-Site: none with Sec-Fetch-Mode: no-cors and Sec-Fetch-Dest: empty

Safari doesn't natively consume OpenSearch suggestions, so it's a non-issue there

That means the gate has to allow both "header absent" and "none" / "same-origin" / "same-site", and only reject explicit cross-site

`Sec-Fetch-*` is on the forbidden header list, so an attacker page can't spoof either the value or its absence; a `fetch()` from `evil.example` will always carry `Sec-Fetch-Site: cross-site`

```go
site := r.Header.Get("Sec-Fetch-Site")
switch site {
case "", "none", "same-origin", "same-site":
    // allow
default:
    http.Error(w, "forbidden", http.StatusForbidden)
    return
}
```

On the per-user token point, you're right that putting a token in the descriptor URL just relocates the same problem to `/opensearch.xml`

Worth separating the two threats though: the cross-user leak utafrali flagged (`UserID == 0` falling through to a global query) is a server-side bug in `serveSuggest` _regardless_ of CSRF, and `Sec-Fetch-Site` _**only**_ addresses the cross-site read of one user's suggestions by an attacker page

asciimoo: Great, thank you for the tests. According to your findings your suggested `Sec-Fetch-Site` solution seems valid. Could you update the PR?

>  the cross-user leak utafrali flagged (UserID == 0 falling through to a global query) is a server-side bug worth fixing in serveSuggest regardless of CSRF

It isn't an issue anymore because you've added the endpoint to the API endpoints where user check is automatic if `NoAuth: true` isn't specified.

4evy: Sorry, I was juggling PRs so I tripped up a bit lol, I'll update the PR

asciimoo (approved):
Thanks!

PR #339, opened 2026-04-15
[ci] add nix flake check workflow

FlameFlag:


asciimoo:
This is introduces a github action for a very rare edge case that runs on every PR. It is more likely to get false positive fails because of github actions instabilities than useful outputs.

PR #341, opened 2026-04-15
[ci] add actionlint workflow and harden base_ref usage

FlameFlag:


asciimoo:
Don't overdo the github actions usage. It isn't the most reliable and can cause a lot of noise in case it isn't available.

PR #343, opened 2026-04-15
[enh] add JSON-LD metadata extractor

FlameFlag:
Parses `<script type="application/ld+json">` tags in fetched HTML and writes normalized schema.org fields (type, headline, description, author, published, modified, image) to `d.Metadata`

Runs first in the chain and always returns Continue so Readability still populates body text. No extra HTTP requests


asciimoo:
It would be good to find a way to efficiently find json-ld content without parsing the whole HTML. Also extracted values should be sanitized.

asciimoo (approved):
Nice, thanks!

PR #345, opened 2026-04-15
Surface preview metadata from readability + JSON-LD

FlameFlag:


webui/app/src/routes/+page.svelte
```diff
@@ -105,6 +113,7 @@
   let panelTemplate = $state('');
   let panelTemplateData = $state<any>(null);
   let panelAdded = $state<number | null>(null);
+  let panelMeta = $state<Record<string, any> | null>(null);
```
asciimoo: why do you differentiate the panel and the popup metadata? Can't we just call it `previewMeta` and use it from both the popup and the panel?

server/server.go
```diff
@@ -968,12 +968,42 @@ func servePreview(c *webContext) {
...
+		payload["meta"] = meta
+	}
+	c.JSON(payload)
+}
+
+// buildPreviewMeta surfaces the document's normalized metadata (merged
+// readability + JSON-LD output) plus the raw JSON-LD node dump so the
+// preview panel can render a byline, description, and an "Extracted
+// JSON-LD" section. Returns nil when there is nothing to surface.
+func buildPreviewMeta(doc *document.Document) map[string]any {
```
asciimoo: Wouldn't be better to make this globally available and to add the function to the `Document` object as something like `GetPreviewMeta() map[string]any`?

FlameFlag:
Thanks for the review, I wrote this PR when it was late night time for me, so not at _my best capacity_

asciimoo (approved):
Nice, thanks!

PR #346, opened 2026-04-15
[tst] add nix module eval checks and CI workflow

FlameFlag:


asciimoo:
I don't think that it is our projects responsibility to provide a complete packaging infrastructure for a specific packaging system, especially while the project is small and we have limited resources. Also the PR seems like having lots of duplicated code/config.
I appreciate nix, but adding hundreds or thousands of lines of code/config to a repo to make it work isn't a good directions. And I hope the nix community agrees on this.
I have other projects that are in the nix package repository (searx, wuzz) and neither of them require any kind of nix configuration in their repo.
I'm more and more feeling inclined to exclude nix from the repo and leave the packaging to the nix community.

- Why do we need this whole nix configuration monster in the repo?
- What are the added benefits of having it in our repo?
- What currently existing problems does this PR solve?

FlameFlag:
> I don't think that it is our projects responsibility to provide a complete packaging infrastructure for a specific packaging system, especially while the project is small and we have limited resources.

To clarify what this PR actually does: it isn't adding packaging infrastructure. It's adding CI tests for the NixOS, nix-darwin, and home-manager modules that already exist in the repo. Without these, those modules can silently break when we change how `dataDir`, `configPath`, `settings`, etc. are handled, and nobody notices until a user hits it. The tests don't get in the way of anything and aren't required to build hister.

On the limited-resources side: the Nix side is self-contained and doesn't touch the build path, so it doesn't pull anyone outside the Nix integration into maintaining it. I'm happy to be on the hook for that part. The tests actually reduce the burden on non-Nix contributors rather than adding to it: if a change to hister accidentally breaks the module, CI will flag it, instead of you or another contributor having to hold "did I just break Nix" in your head for every PR

> I appreciate nix, but adding hundreds or thousands of lines of code/config to a repo to make it work isn't a good direction. And I hope the nix community agrees on this.

The core Nix integration really only needs `flake.nix` and `flake.lock`. Most of the new lines in this PR are the test harnesses, which are short grep-based checks against the rendered systemd units and launchd plists. That's the same pattern nixpkgs, nix-darwin, and home-manager use in their own test suites, so we're not inventing anything custom here.

It's also mostly a one-time cost. The systemd, launchd, and home-manager module APIs are very stable, and the assertions in the tests are anchored on primitives like `User=`, `ProgramArguments`, and `CAP_NET_BIND_SERVICE` that haven't moved in years. Once this lands, the tests only really need touching when we add or change one of hister's own config options, which is something you'd want covered anyway. The line count is front-loaded, _not a recurring tax_.

> I have other projects that are in the nix package repository (searx, wuzz) and neither of them require any kind of nix configuration in their repo.

Right, a project doesn't need anything in-repo to end up in nixpkgs. What's in the repo here exists so Nix users can consume hister directly via the flake with a ready-to-use module, without waiting for nixpkgs release cycles or having to write their own. I've tried to make that experience as rich as possible on our side.

> I'm more and more feeling inclined to exclude nix from the repo and leave the packaging to the nix community.

That's fair. I think hister has existed long enough and has enough traction to be a reasonable candidate for nixpkgs at this point, so I'll look into upstreaming it (with the NixOS module, and ideally home-manager too). Once that's done we can drop our in-repo module system entirely, and the repo would only really need `flake.nix`/`flake.lock` as a convenience for Nix users, with nothing else for us to maintain.

> Also the PR seems like having lots of duplicated code/config.

The duplication inside `nix/checks/` is structural rather than copy-paste. Each of the three module systems has a different entry point (`nixosSystem`, `darwinSystem`, `homeManagerConfiguration`) and produces a different kind of artifact: NixOS renders a systemd unit as text, nix-darwin renders a launchd plist file, and home-manager builds an activation package containing one or the other depending on platform. The properties worth checking also differ per target. NixOS is where `User`/`Group`/`CAP_NET_BIND_SERVICE`/firewall rules live, launchd cares about `ProgramArguments` and `WorkingDirectory`, and home-manager runs at user level so neither applies cleanly. Collapsing all three into one shared helper would mean inventing a custom abstraction layer that hides what's actually being asserted, which tends to make breakages harder to diagnose rather than easier. Keeping one small file per target, each following the idiom that nixpkgs, nix-darwin, and home-manager use in their own tests, felt like the clearer choice.

---

To answer your questions more directly:

> Why do we need this whole nix configuration monster in the repo?

We don't strictly need it. The hard minimum is `flake.nix`/`flake.lock`. Everything else exists to give Nix users a ready-to-use, tested deployment path rather than asking them to write their own module or wait for nixpkgs to catch up.

> What are the added benefits of having it in our repo?

Users on NixOS, nix-darwin, or home-manager can deploy hister with `services.hister.enable = true;` and a few options (`port`, `dataDir`, `settings`, `environmentFile`, etc.), instead of rolling their own systemd unit or launchd plist. Those options map to hardened, idiomatic service definitions, including systemd sandboxing flags (`NoNewPrivileges`, `ProtectSystem=strict`, `MemoryDenyWriteExecute`), auto-created system user/group, and `CAP_NET_BIND_SERVICE` plus firewall handling for privileged ports. New functionality is also available to Nix users on the same release cadence as the rest of the project, rather than being gated on nixpkgs staging, which can be weeks behind.

> What currently existing problems does this PR solve?

Until now the three modules lived in the repo but had no automated coverage. A change to how `settings` is rendered, where `HISTER_DATA_DIR` points, or whether `CAP_NET_BIND_SERVICE` is granted for privileged ports could silently break the module, and we'd only find out when a user reported a broken deployment. This PR adds a real NixOS VM test that boots hister and hits it with curl (catching integration regressions), unit-level assertions against the rendered systemd/launchd output (catching smaller regressions without a full VM boot), and parity coverage for nix-darwin and home-manager so a fix on one platform doesn't quietly regress the other two.

asciimoo:
Thanks for the detailed explanation

> On the limited-resources side: the Nix side is self-contained and doesn't touch the build path,

We already have 54 nix related commits (`git log --pretty=format:"%s" | grep -i nix | wc -l`). Every 19-20th commit is nix related on average. This consumes time, distracts attention and creates big noise for everybody who follows the project while we have zero user requests asking for nix support or make it more sophisticated.

> That's fair. I think hister has existed long enough and has enough traction to be a reasonable candidate for nixpkgs at this point,

Great, then I'd suggest to start transitioning to this direction instead of adding more and more nix related things to this repo.

FlameFlag:
> We already have 54 nix related commits

But most of those are automated `[chore]` hash bumps and flake.lock updates rather than substantive changes

Filtering them out with `git log --pretty=format:"%s" | grep -i nix | grep -vE "^\[chore\]|^Merge|^Bump"` gives 13 commits, and one of even those is a false positive (`grep -i nix` also matches "unix")

> Great, then I'd suggest to start transitioning to this direction instead of adding more and more nix related things to this repo.

In that case, I'll close the PR and I'll start looking into upstreaming hister to nixpkgs with the NixOS module, once those are in, we can strip what's in this repo down to just `flake.nix`/`flake.lock`

Thanks for your time in any case

PR #350, opened 2026-04-16
[enh] add Wikipedia extractor, extract shared urlutil, extend SVG sanitizer

FlameFlag:


server/sanitizer/sanitizer.go
```diff
@@ -12,82 +13,125 @@ var (
...
-		"color",
-		"background",
-		"background-color",
-		"background-image",
-		"font-size",
-		"text-align",
+		// Text.
+		"text-decoration", "color", "font-size", "font-weight", "font-style", "text-align",
+		// Background.
+		"background", "background-color", "background-image",
```
asciimoo: Rules that can embed third party content should not be allowed

server/sanitizer/sanitizer.go
```diff
@@ -12,82 +13,125 @@ var (
...
 	).Globally()
+
 	p.AllowAttrs("href").OnElements("a")
 	p.AllowAttrs("src", "srcset").OnElements("img", "source")
-	p.AllowAttrs(
-		"alt",
-		"title",
-		"id",
-	).Globally()
+	p.AllowAttrs("alt", "title", "id").Globally()
```
asciimoo: id should not be allowed globally, since it can conflict with the page

webui/app/src/style.css
```diff
@@ -80,4 +80,17 @@
...
+  /* Preview panel: allow wide tables to scroll horizontally and prevent
+     floated elements (infoboxes, thumbnails) from overflowing. */
+  .prose table {
+    overflow-x: auto;
+    max-width: 100%;
+  }
+
+  .prose::after {
+    content: '';
+    display: table;
```
asciimoo: `display: table`? why?

server/extractor/extractors/wikipedia/wikipedia.go
```diff
@@ -0,0 +1,249 @@
...
+	content := articleContent(doc)
+	if content.Length() == 0 {
+		return types.ExtractorContinue, fmt.Errorf("no mw-parser-output found")
+	}
+
+	if d.Metadata == nil {
+		d.Metadata = make(map[string]any)
+	}
+	d.Metadata["type"] = "Article"
+	d.Metadata["site_name"] = "Wikipedia"
```
asciimoo: What is the point of storing this? We know from the URL that this is wikipedia

server/extractor/extractors/wikipedia/url.go
```diff
@@ -0,0 +1,46 @@
+package wikipedia
```
asciimoo: The functions of this file could be part of the `urlutils`

server/extractor/extractors/wikipedia/style.go
```diff
@@ -0,0 +1,333 @@
+package wikipedia
```
asciimoo, line 1: Wouldn't be better to create a wikipedia/general wiki template in the UI instead of string concatenating formatted HTML in go?

Then the wikipedia extractor could pass a JSON content to the UI and the UI can do the formatting.

4evy: Yes, but I think that's out of the scope of this PR and the current implementation here is a good starting point

asciimoo: This approach here feels like more of a double effort than an incremental development, because this whole html generator solution should be replaced later on with the JSON/Svelte combo.
What is the benefit of doing it in two steps?

4evy: Most of the work is reusable and designing the JSON schema is easier once we know what we're rendering

The bulk of this PR is the `goquery` parsing (`infoboxes`, `hatnotes`, `thumbs`, references, cleaning up editorial cruft, etc.), and that logic stays when we switch to JSON, it just emits a structured tree instead of HTML strings. 

The only throwaway part is the inline style constants in `style.go`.

Shipping the HTML version first also surfaces the edge cases on real articles (nested infoboxes, math, galleries, navboxes, ...), so the Svelte contract can be designed against something concrete instead of guessed upfront

Scope-wise, the SVG `sanitizer` and `urlutil` changes are useful on their own regardless of how Wikipedia output is eventually rendered, and bundling a frontend template system on top would make this PR quite a bit harder to review imo

asciimoo: Alright. I still believe, that this isn't the ideal way, but the extractor is working and adds value to the project, so I'm merging it. Thanks for developing it.

server/sanitizer/sanitizer.go
```diff
@@ -12,82 +13,134 @@ var (
...
-		"text-align",
+		// Box model.
+		"border", "border-top", "border-bottom", "border-left", "border-right",
+		"border-collapse", "border-spacing", "border-color", "border-radius",
+		"padding", "padding-top", "padding-bottom", "padding-left", "padding-right",
+		"margin", "margin-top", "margin-bottom", "margin-left", "margin-right",
+		// Sizing.
+		"width", "max-width", "min-width", "height", "max-height",
+		// Layout.
+		"float", "clear", "display", "position",
```
asciimoo: I'd remove display, position, float, top, left, right, bottom values because it allows malicious content to overlap with the original page content. An attacker can craft a fake URL bar image and position it where our original URL bar is and if the user clicks on it the link redirects them to a malicious website.

asciimoo (approved):
Awesome, thanks!

PR #352, opened 2026-04-18
Add vector/similarity search

asciimoo:
This PR introduces optional semantic (vector similarity) search alongside the existing full-text search. When enabled, documents are automatically embedded and results from both engines are merged and re-ranked.

When a document is indexed, its text is split into overlapping token chunks and each chunk is sent to a configurable OpenAI-compatible embeddings endpoint. The resulting vectors are stored in a separate database (SQLite via sqlite-vec, or PostgreSQL via pgvector). On search, the query is also embedded and the top-K most similar chunks are retrieved, then merged with keyword results using a configurable weight.

Rather than depending on github.com/asg017/sqlite-vec-go-bindings/cgo (which requires a system libsqlite3-dev at compile time, not available for cross-compilation targets AFAIK), sqlite-vec.c, sqlite-vec.h, and sqlite3.h (from mattn/go-sqlite3) are bundled directly in the repo. This change allows us to not introduce extra external dependency and keeps the existing goreleaser/goreleaser-cross CI setup working unchanged for all targets. An update.sh script makes it easy to refresh the bundled files when either upstream dependency is upgraded.

Example config:

```yaml
semantic_search:
  enable: true
  embedding_endpoint: http://localhost:11434/v1/embeddings
  embedding_model: nomic-embed-text
  dimensions: 768
  max_context_length: 512
  chunk_overlap: 50
  similarity_threshold: 0.5
  result_limit: 10
  semantic_weight: 0.4
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -126,15 +126,17 @@ type Extractor struct {
 
 // SemanticSearch holds configuration for optional vector similarity search.
 type SemanticSearch struct {
-	Enable              bool    `yaml:"enable" mapstructure:"enable"`
-	EmbeddingEndpoint   string  `yaml:"embedding_endpoint" mapstructure:"embedding_endpoint"`
-	EmbeddingModel      string  `yaml:"embedding_model" mapstructure:"embedding_model"`
-	Dimensions          int     `yaml:"dimensions" mapstructure:"dimensions"`
-	MaxContextLength    int     `yaml:"max_context_length" mapstructure:"max_context_length"`
-	ChunkOverlap        int     `yaml:"chunk_overlap" mapstructure:"chunk_overlap"`
-	SimilarityThreshold float64 `yaml:"similarity_threshold" mapstructure:"similarity_threshold"`
-	ResultLimit         int     `yaml:"result_limit" mapstructure:"result_limit"`
-	SemanticWeight      float64 `yaml:"semantic_weight" mapstructure:"semantic_weight"`
+	Enable              bool              `yaml:"enable" mapstructure:"enable"`
+	EmbeddingEndpoint   string            `yaml:"embedding_endpoint" mapstructure:"embedding_endpoint"`
+	EmbeddingModel      string            `yaml:"embedding_model" mapstructure:"embedding_model"`
+	APIKey              string            `yaml:"api_key" mapstructure:"api_key"`
+	Headers             map[string]string `yaml:"headers" mapstructure:"headers"`
+	Dimensions          int               `yaml:"dimensions" mapstructure:"dimensions"`
+	MaxContextLength    int               `yaml:"max_context_length" mapstructure:"max_context_length"`
+	ChunkOverlap        int               `yaml:"chunk_overlap" mapstructure:"chunk_overlap"`
+	SimilarityThreshold float64           `yaml:"similarity_threshold" mapstructure:"similarity_threshold"`
+	ResultLimit         int               `yaml:"result_limit" mapstructure:"result_limit"`
+	SemanticWeight      float64           `yaml:"semantic_weight" mapstructure:"semantic_weight"`
 }
 
 // DBTypedef represents the type of database being used.
--- a/config/config.go
+++ b/config/config.go
@@ -34,6 +34,7 @@ type Config struct {
 	Server                   Server                `yaml:"server" mapstructure:"server"`
 	Indexer                  Indexer               `yaml:"indexer" mapstructure:"indexer"`
 	Crawler                  CrawlerConfig         `yaml:"crawler" mapstructure:"crawler"`
+	SemanticSearch           SemanticSearch        `yaml:"semantic_search" mapstructure:"semantic_search"`
 	Hotkeys                  Hotkeys               `yaml:"hotkeys" mapstructure:"hotkeys"`
 	TUI                      TUI                   `yaml:"-" mapstructure:"tui"`
 	SensitiveContentPatterns map[string]string     `yaml:"sensitive_content_patterns" mapstructure:"sensitive_content_patterns"`
@@ -121,6 +122,17 @@ type Rule struct {
 type Extractor struct {
 	Enable  bool           `yaml:"enable" mapstructure:"enable"`
 	Options map[string]any `yaml:"options" mapstructure:"options"`
+}
+
+// SemanticSearch holds configuration for optional vector similarity search.
+type SemanticSearch struct {
+	Enable              bool    `yaml:"enable" mapstructure:"enable"`
+	EmbeddingEndpoint   string  `yaml:"embedding_endpoint" mapstructure:"embedding_endpoint"`
+	EmbeddingModel      string  `yaml:"embedding_model" mapstructure:"embedding_model"`
+	Dimensions          int     `yaml:"dimensions" mapstructure:"dimensions"`
+	SimilarityThreshold float64 `yaml:"similarity_threshold" mapstructure:"similarity_threshold"`
+	ResultLimit         int     `yaml:"result_limit" mapstructure:"result_limit"`
+	SemanticWeight      float64 `yaml:"semantic_weight" mapstructure:"semantic_weight"`
 }
 
 // DBTypedef represents the type of database being used.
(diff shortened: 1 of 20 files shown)


FlameFlag:
One thing worth noting: we could use https://github.com/ncruces/go-sqlite3 which is a pure-Go SQLite driver (SQLite compiled to WASM, run via wazero), no CGO, cross-compiles to every target with stock go build, and has first-class sqlite-vec support as a WASM extension

That would let us drop the vendored `sqlite-vec.c` / `sqlite-vec.h` / `sqlite3.h` shim and the `update.sh` dance entirely, since there's nothing to compile at the C level

asciimoo:
Yes, I've checked it as a potential alternative, however the project does not seem mature/stable enough to me. There are some recently closed issues about failing builds, pointer errors. Also, the documentation mentions that it has inferior performance compared to the C solution, not entirely sure what this means in reality.

As I understand it is a 1:1 replacement to the C based driver, so we can switch any time later on if the benefits are clear and the project seems stable.

server/vectorstore/embedder.go
```diff
@@ -0,0 +1,161 @@
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("marshal embedding request: %w", err)
+	}
+
+	req, err := http.NewRequest("POST", e.endpoint, bytes.NewReader(body))
+	if err != nil {
+		return nil, fmt.Errorf("create embedding request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
```
4evy, line 75: No `Authorization` header or API-key support. This works against a local Ollama or llama.cpp endpoint, but any hosted OpenAI-compatible service (OpenAI itself, Together, Mistral, Voyage via gateway, Azure, etc.) requires `Authorization: Bearer ...`.

There's also no corresponding `api_key` field in the `SemanticSearch` config struct, so there's no way to plumb one in without code changes.

Consider adding an optional `api_key` field (and maybe generic `headers`) and setting the header here. The README's "OpenAI-compatible endpoint" framing implies hosted providers work, but today only un-authed local ones do.

server/indexer/indexer.go
```diff
@@ -202,6 +232,21 @@ func Reindex(basePath string, rules *config.Rules, skipSensitiveChecks bool, det
...
+	// separate file from the Bleve indexes).
+	vs := idx.vectorStore
+	embedder := idx.embedder
+	if vs != nil && embedder != nil {
+		if err := vs.Clear(); err != nil {
+			log.Warn().Err(err).Msg("failed to clear vector store before reindex")
+		} else {
+			tmpIdx.vectorStore = vs
+			tmpIdx.embedder = embedder
+		}
```
4evy, line 248: The Bleve reindex and the vector store handle failures asymmetrically.

Bleve: the new indexes are built under `basePath/reindex`, and the old indexes are only removed once the rebuild loop completes successfully. A mid-loop failure leaves the old indexes untouched, so keyword search keeps working with the pre-reindex data.

Vector store: `vs.Clear()` wipes every embedding *before* the loop runs. A mid-loop failure (embedding endpoint down, panic, SIGTERM, a single embed error returning `err` from `b.Add`, etc.) leaves the vector DB empty with no way back except another full reindex.

The inline comment ("no temp-dir / rename dance is needed because it is a separate file") justifies the in-place approach, but being a separate file arguably makes temp/rename *easier*, not unnecessary. Worth mirroring the Bleve approach, or at least calling this out in the docs so operators understand a failed reindex equals an empty vector DB.

server/indexer/indexer.go
```diff
@@ -350,6 +441,11 @@ func (i *indexer) AddDocument(d *document.Document) error {
 			return err
 		}
 	}
+	if i.embedder != nil && i.vectorStore != nil {
+		i.embedWg.Go(func() {
+			embedDocumentChunks(i, d)
+		})
+	}
 	return i.getOrCreate(d.Language).Index(d.ID(), d)
```
4evy, line 448: Async embedding is fire-and-forget: no context, no retry, errors only logged. A single transient failure from the embedding endpoint (ollama restart, rate limit, network blip) silently drops that document's vectors forever. Semantic search will never return it, and there's no reconciliation pass to backfill.

Two small follow-ups worth considering:
1. Thread a `context.Context` through `Embedder` (use `http.NewRequestWithContext`) so `Close()` can cancel in-flight requests instead of blocking up to ~30s per goroutine on `embedWg.Wait()` at shutdown.
2. Either retry transient failures, or persist a "needs embedding" marker so a background job can backfill.

config/config.go
```diff
@@ -123,6 +124,19 @@ type Extractor struct {
...
+type SemanticSearch struct {
+	Enable              bool    `yaml:"enable" mapstructure:"enable"`
+	EmbeddingEndpoint   string  `yaml:"embedding_endpoint" mapstructure:"embedding_endpoint"`
+	EmbeddingModel      string  `yaml:"embedding_model" mapstructure:"embedding_model"`
+	Dimensions          int     `yaml:"dimensions" mapstructure:"dimensions"`
+	MaxContextLength    int     `yaml:"max_context_length" mapstructure:"max_context_length"`
+	ChunkOverlap        int     `yaml:"chunk_overlap" mapstructure:"chunk_overlap"`
+	SimilarityThreshold float64 `yaml:"similarity_threshold" mapstructure:"similarity_threshold"`
+	ResultLimit         int     `yaml:"result_limit" mapstructure:"result_limit"`
+	SemanticWeight      float64 `yaml:"semantic_weight" mapstructure:"semantic_weight"`
```
4evy: No validation when `semantic_search.enable: true`. If `embedding_endpoint` is blank, `dimensions` is 0, etc., the user hits confusing runtime errors (dimension mismatch on every embed, POST to `""`, `CREATE TABLE ... vector(0)` which errors at insert time).

A startup check with a clear message, in the spirit of `Hotkeys.Validate()` elsewhere in this file, would save a lot of head-scratching.

server/vectorstore/tokenizer.go
```diff
@@ -0,0 +1,84 @@
...
+
+// tokenize splits text into tokens on whitespace and punctuation boundaries.
+// Each contiguous run of letters/digits is a token. This is a simple
+// approximation — it doesn't match any specific model's BPE tokenizer, but is
+// good enough for determining chunk boundaries.
+func tokenize(text string) []string {
+	var tokens []string
+	var cur strings.Builder
+	for _, r := range text {
+		if unicode.IsLetter(r) || unicode.IsDigit(r) {
```
4evy: `unicode.IsLetter` is true for CJK ideographs, Thai, Lao, etc., which are scripts that don't use whitespace separators. I confirmed this locally: a 51-rune Chinese sentence tokenizes to exactly **1 token**.

A long Chinese or Japanese document therefore returns a single chunk whose token count satisfies `<= maxTokens`, but whose rune count vastly exceeds the embedding model's real context window. The endpoint then silently truncates or errors.

Not a blocker tho, worth either a commenting the limitation, or a rune-count budget fallback for non-space-separated scripts (e.g. treat each ideograph as a token for chunk-budgeting purposes).

server/vectorstore/postgres.go
```diff
@@ -0,0 +1,162 @@
...
+		if i > 0 {
+			sb.WriteString(", ")
+		}
+		base := i*cols + 1
+		fmt.Fprintf(&sb, "($%d, $%d, $%d, $%d, $%d, $%d)", base, base+1, base+2, base+3, base+4, base+5)
+		args = append(args, chunkKey(docID, c.Index), docID, c.Index, userID, c.Text, pgVectorLiteral(c.Embedding))
+	}
+	if _, err = tx.Exec(sb.String(), args...); err != nil {
+		return fmt.Errorf("insert embedding chunks: %w", err)
+	}
```
4evy, line 100: Minor, defensive: `PutChunks` with `len(chunks) == 0` would produce malformed SQL (`INSERT ... VALUES ` with nothing after).

The only current caller (`embedDocumentChunks`) early-returns on empty, so it's safe today, but a one-line guard here would make the invariant local to this function.

FlameFlag (changes requested):

asciimoo:
@FlameFlag Thanks a lot for the review. I agree with all the findings. I'll either fix them or open an issue to do it in the future if the note is minor.

server/indexer/indexer.go
asciimoo: That's true, long term it would be good to make the semantic search reindex as fail safe as the normal indexer's reindex. I'd leave the resolution of this issue to the future, because it is complex and it would significantly increase the size of this PR (even more)..

4evy: Sure

PR #364, opened 2026-04-22
[enh] indexer: add facets, MatchAll, and user-scoped lookup

FlameFlag:
Lays the groundwork for filter sidebars and smarter autocomplete on search

Search results can now come back with aggregate counts: which domains show up
most in your history, which languages, and how results break down by recency
(last 24h, 7 days, 30 days, year, older). That's what a sidebar needs to show
"GitHub (42), Wikipedia (17), ..." next to the results, or to power a recency
filter without a second round-trip

Autocomplete and filter UIs can also ask for these counts on their own, without
running a text search, so pulling the list of domains a user has visited is
cheap

Fixes a cross-user bug in URL lookup: in multi-user mode, looking up a URL could
return another user's copy of the same page. Callers can now scope the lookup to
the current user

### DIFF
--- a/server/indexer/indexer.go
+++ b/server/indexer/indexer.go
@@ -886,13 +886,7 @@ func Search(cfg *config.Config, q *Query) (*Results, error) {
 						MatchedChunk: truncateText(dh.chunkText, semanticTextPreviewLen),
 					}
 					// For semantic-only hits, populate the document with a truncated text preview.
-					// vectorStore.Search was scoped to q.UserID, so every docID
-					// here has that uid prefix (see document.GetDocID).
-					url := docID
-					if q.UserID > 0 {
-						url = strings.TrimPrefix(docID, fmt.Sprintf("%d:", q.UserID))
-					}
-					d := GetByURLAndUser(url, q.UserID)
+					d := GetByDocID(docID)
 					if d != nil {
 						if _, inKeyword := keywordURLs[d.URL]; !inKeyword {
 							d.Text = truncateText(d.Text, semanticTextPreviewLen)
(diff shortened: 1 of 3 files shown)


asciimoo:
Nice idea!

> That's what a sidebar needs to show "GitHub (42), Wikipedia (17), ..." 

How do you imagine displaying these information? Adding an extra sidebar doesn't make the search UI too dense?

> Fixes a cross-user bug in URL lookup

This is a bug indeed, but you don't actually provide a solution for it, only tools that can solve the issue. `GetByURLAndUser` isn't used anywhere. In my opinion it would be good to fully replace `GetByURL` with `GetByURLAndUser` and use it everywhere instead of `GetByURL`.

FlameFlag:
> How do you imagine displaying these information? Adding an extra sidebar doesn't make the search UI too dense?

This PR is only the backend groundwork so the frontend has something to call

But I was thinking:

- Collapsible sidebar (hidden by default on narrow viewports, toggled from the header)
- A compact filter bar above the results (chips like github.com · 42 that toggle as filters) instead of a full sidebar
- Facets only surfaced on an explicit "filters" view, with autocomplete consuming the same endpoint silently

I lean toward the chip-bar for the density reason you mentioned, it keeps the main column unchanged when no filter is active, but I think this is for another PR

> This is a bug indeed, but you don't actually provide a solution for it, only tools that can solve the issue. GetByURLAndUser isn't used anywhere. In my opinion it would be good to fully replace GetByURL with GetByURLAndUser and use it everywhere instead of GetByURL.

Ah, right I'll fix it

server/indexer/indexer.go
```diff
@@ -807,9 +908,20 @@ func Search(cfg *config.Config, q *Query) (*Results, error) {
...
-	q.SetField("url")
+// GetByURLAndUser returns the document at u owned by uid. The url field is
+// shared across owners in multi-user mode, so callers must pass their own
+// UserID to avoid returning another user's copy of the same URL. A uid of 0
+// matches the global (single-user) owner.
+func GetByURLAndUser(u string, uid uint) *document.Document {
+	urlQ := query.NewTermQuery(strings.ToLower(u))
+	urlQ.SetField("url")
+	var q query.Query = urlQ
+	if uid > 0 {
```
asciimoo: I think we should set the user ID filtering even if it is 0, just in case an instance have publicly available documents with user id 0 and private user documents as well.

Perhaps an even better solution is to use `document.GetDocID(uid, u)` function and retrieve the document by the provided document/index ID instead of filtering for `document.URL + document.UserID` combo.

asciimoo (approved):
Nice, thanks!

PR #365, opened 2026-04-22
[refactor] webui: extract SiteHeader and SiteFooter from layout

FlameFlag:
Split the layout's header and footer into dedicated components and
collapse the repeated footer-link boilerplate into a data-driven list


asciimoo (approved):
Nice, thanks!

PR #389, opened 2026-05-04
[enh] add multi-language subtitle support to yt-dlp extractor

dummylabs:
First of all, I really like your project; I use it actively and will follow its development with interest.

# Disclaimer

I am not a golang expert; however, I have sufficient experience in other programming languages, so I used Claude Code for AI-assisted development, not vibe-coding. This text was written by a human and verified by an LLM for mistakes.

# Problem Statement

If the `sub_language` setting in the ytdlp extractor is set to a language that does not match the video's default language, the video transcript is not saved in Hister because yt-dlp returns 201 and an empty string.

A workaround is to request an auto-translation of captions into the specified language, but in this case YouTube applies stricter checks, and cookies may be required, overwise it may return `429: Too Many Requests`. Therefore, in my experiments, the most reliable approach (without extra impersonation or cookie import) was to request captions in the language that matches the video's default language (`info.Language`).

# Core Changes

In the ytdlp extractor configuration, the `sub_language` parameter can now take the value `auto`, or contain one or more subtitle languages separated by commas. 
Logic:
- if `auto` is specified: download subtitles in the default language for this video
- if `en` is specified: download subtitles in English or nothing
- if `en`, `fr` are specified: try downloading subtitles in the listed languages in order of priority, or nothing if none of the specified languages is the video's default.

If `en` is specified as `sub_language`, the current behavior (before the change) is fully preserved.

# Changes to the default configuration

```yaml
extractors:
    ytdlp:
        fetch_subtitles: true
        sub_language: "auto"      # detect original language automatically
```

# Testing
Tested manually on videos with different languages and different values of `sub_language`.  

Added new test fixtures:
- `sampleJSONTemplateFr` for a non-English language test
- `sampleJSONTemplateNoLang` for a video without a default language (this is the real case in my experience)

Four new tests added:
  
  |       Test       | Fixture | sub_language |    Expectation     |
  |------------|--------|---------------|-----------------|
| Auto             | Fr      | `auto`        | subtitles present  |
   | MultiLangMatch   | Fr      | `fr,en`       | subtitles present  |
  | AutoNoLanguage   | NoLang  | `auto`         | no subtitles       |
  | MultiLangNoMatch | Fr      | `de,es`        | no subtitles       |

### DIFF
--- a/server/extractor/extractors/ytdlp/types.go
+++ b/server/extractor/extractors/ytdlp/types.go
@@ -33,6 +33,7 @@ type videoInfo struct {
 	Chapters          []chapter                  `json:"chapters"`
 	Subtitles         map[string][]subtitleTrack `json:"subtitles"`
 	AutomaticCaptions map[string][]subtitleTrack `json:"automatic_captions"`
+	Language          string                     `json:"language"`
 	PlaylistTitle     string                     `json:"playlist_title"`
 	PlaylistIndex     *int                       `json:"playlist_index"`
 	PlaylistCount     *int                       `json:"playlist_count"`
--- a/server/extractor/extractors/ytdlp/vtt.go
+++ b/server/extractor/extractors/ytdlp/vtt.go
@@ -5,15 +5,46 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"slices"
 	"strings"
 	"unicode"
 )
 
-// fetchSubtitleText uses yt-dlp to download subtitles and strips timing/formatting
-// to return plain transcript text. It prefers manual subtitles over auto-captions.
+// fetchSubtitleText dispatches subtitle downloading based on the sub_language config:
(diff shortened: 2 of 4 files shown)


asciimoo:
Thank you both for the contribution and for being transparent about AI usage.

The pull requests looks good to me.

I just realized that we do not document extractor specific configuration options anywhere. Would you be open to add `yt-dlp` config options to the extractor documentation (`webui/website/src/content/docs/extractors.md`) in a follow-up pull request?

asciimoo:
Please fix the go lint error and it's ready to merge

asciimoo:
@dummylabs doh, sorry the lint error is not even in your code.. My fault, I'm fixing it. Not sure how my local linter let that slip..

asciimoo (approved):
Thanks!

PR #396, opened 2026-05-05
bidi backend, fix #284

### LINKED ISSUE #284: more flexible crawling browser choice
Currently Chromium-specific. But Firefox could be the UA holding all the cookies/session/localStorage/...

Puppeteer (or a similar Go library) would offer more possibilities to interact with different browsers (headless or not).

JS way to handle it (here for Firefox) is specifying a WS session endpoint:
```js
    const browser = await puppeteer.connect({
        browserWSEndpoint: 'ws://127.0.0.1:9123/session',
        protocol: 'webDriverBiDi',
        browser: 'firefox'
    });
```

Something around the lines in the YAML configuration would allow to use any (debugger-enabled) browser so that state/cookie/session are implicitly decided by the user based on debugger configuration (and debugger port selection).
This allows for multiple browser instances on distinct ports which paves the way for "per-domain/per-pattern crawling browsers" .

drzraf:
Successfully tested with Firefox

Happened that hupe1980/gowebdriver had problems regarding leftovers sessions (closing sessions) that would lead to quickly reach browser's BiDi "max sessions". Since websocket was already pulled, it wasn't _that_ much of an additional to internalize the `session.new/browsingContext.navigate/script.evaluate/session.end` flow.

fix #284

(opus 4.6)

### DIFF
--- a/hister.go
+++ b/hister.go
@@ -680,7 +680,7 @@ func init() {
 	indexCmd.Flags().Bool("global", false, "Make indexed documents available for all users (only for admins in multiuser mode)")
 	indexCmd.Flags().Uint("user-id", 0, "Index documents under the given user ID (only for admins in multiuser mode)")
 	indexCmd.Flags().String("job-id", "", "Persistent crawl job ID; use with --recursive to start a new job or alone to resume an existing one")
-	indexCmd.Flags().String("backend", "", "Crawler backend to use (\"http\" or \"chromedp\")")
+	indexCmd.Flags().String("backend", "", "Crawler backend to use (\"http\", \"chromedp\", or \"bidi\")")
 	indexCmd.Flags().StringToString("backend-option", nil, "Crawler backend option as key=value (repeatable, e.g. --backend-option exec_path=/usr/bin/chromium)")
 	indexCmd.Flags().StringToString("header", nil, "Extra HTTP header as KEY=VALUE (repeatable, e.g. --header Accept-Language=en)")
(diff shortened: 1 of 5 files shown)


asciimoo:
Great addition, thanks! Please fix the formatting/linter errors. NPM formatting can be fixed by running `npm run format`.

asciimoo:
Seems like the formatting commits made thing worse.  use [`golangci-lint run -v ./...`](https://golangci-lint.run/docs/) to check go linting locally

asciimoo (approved):
Awesome, thanks!

drzraf:
~~Just for future guidance, regarding the way you merge, I'd personally prefer merges to happen the "regular Github way" which in the UI end up referencing only the usernames instead of resulting with full name expansion.
thx~~

asciimoo:
> Just for future guidance, regarding the way you merge.

I prefer merging this way, because in the git log tree view it is cleary shown what commits belong to a single PR. If you prefer to not display your full name, you should change it in your local git config, because otherwise either github displays it or not, it is going to end up in the git log anyway.

drzraf:
~~Sure, but aside from strict PII aspects, it's about the merging strategy respecting what users set in their GitHub preferences.~~

asciimoo:
>  it's about the merging strategy respecting what users set in their GitHub preferences.

Wouldn't it lead to an inconsistent project git log? In my experience projects use their own preferred merging strategies instead of the contributor's, no?

I've never seen anybody asking to use their preferred merging strategy before - didn't even know it is a thing. Do I miss something?

drzraf:
Oh, my bad. **I** messed-up with email addresses in my GH profile. Just forget about this. Sorry for the noise.

PR #397, opened 2026-05-05
avoid duplicated session when indexing multiple URLs

drzraf:
Related to https://github.com/asciimoo/hister/pull/396
but a separated diff/PR won't hurt.

indexURL() is called per-URL with plain `index URLa URLb`.
Each call creates a brand-new crawler (WebSocket connection + session.new + session.end + close) what is very wasteful for bidi.

=> create the crawler once before the loop and pass it into indexURL


asciimoo (approved):
Thanks!

PR #415, opened 2026-05-17
New extractor: Notion

kvch:
This PR adds a new `Extractor` for Notion pages. It captures text and preview from notion.so and notion.site pages.


asciimoo (approved):
Great, thanks!

PR #419, opened 2026-05-20
[enh] improve embedding request handling

### LINKED ISSUE #355: Add context to async embedding calls
Async embedding is fire-and-forget: no context, no retry, errors only logged. A single transient failure from the embedding endpoint (ollama restart, rate limit, network blip) silently drops that document's vectors forever. Semantic search will never return it, and there's no reconciliation pass to backfill.

Two small follow-ups worth considering:

Thread a context.Context through Embedder (use http.NewRequestWithContext) so Close() can cancel in-flight requests instead of blocking up to ~30s per goroutine on embedWg.Wait() at shutdown.
Either retry transient failures, or persist a "needs embedding" marker so a background job can backfill.

FlameFlag:
Closes #355

### DIFF
--- a/server/indexer/indexer.go
+++ b/server/indexer/indexer.go
@@ -1,6 +1,7 @@
 package indexer
 
 import (
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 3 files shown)


asciimoo (approved):
Nice, thanks!

PR #420, opened 2026-05-20
[enh] update fallow configuration

FlameFlag:


.fallowrc.json
```diff
@@ -10,36 +11,48 @@
...
+    "ignoreImports": true
   },
   "health": {
     "maxCyclomatic": 20,
-    "maxCognitive": 25
+    "maxCognitive": 25,
+    "maxCrap": 30,
+    "suggestInlineSuppression": true
+  },
+  "audit": {
```
4evy, line 47: this is just for now, it ignores any _existing_ flaws for files and only focuses on new stuff, so don't get dragged down but eventually it should be removed

asciimoo:
Have you checked the new version of fallow if it contains anything suspicious/supply chain attack?

4evy:
Such sad times this has to be a concern!

But here is the diff: https://github.com/fallow-rs/fallow/compare/v2.75.0...v2.76.0

The update is mostly a schema/output-contract cleanup and typed JSON-output migration, with a some analyzer behavior improvements

asciimoo (approved):
Thanks!

asciimoo:
> Such sad times this has to be a concern!

It is always good to check the changes before bumping dependencies regardless the increased frequency of these kind of attacks.

PR #426, opened 2026-05-21
improve users' ability to diagnose yt-dlp configuration errors

masukomi:
this commit passes on the error output from yt-dlp rather than just eating it.


asciimoo (approved):
Thanks!

Issue #430, opened 2026-05-25
[feature] Add user to file indexing config


troed:
Example: A self hosted Nextcloud instance can have each user's own notes (markdown files) in separate directories accessable by the Hister server. Such notes folders could then be indexed by Hister where each user only gets access to the notes content from their individual folders.

```
indexer:
  directories:
    - path: '/nextcloud/notes/user1'
      user: user1
      filetypes: ['txt', 'md']
    - path: '/nextcloud/notes/user2'
      user: user2
      filetypes: ['txt', 'md']
```

troed:
If there are no protests I'd be happy to do the work on this

asciimoo:
Definitely no protest, I'd appreciate it. =]

asciimoo:
Added in #438.

PR #438, opened 2026-05-25
Implemented user as optional field on indexer.directories.path

troed:
This PR solves feature suggestion #430 

New field `user`:

```
indexer:
    detect_languages: true
    directories:
      - path: '/notes/troed'
        user: troed
        filetypes: ['md']
        delete_on_remove: true
```

If configured, indexed files under that path are only visible to the corresponding Hister user. Hister will refuse to start if a username is configured that doesn't exist, or if a username is configured but user_handling is not set to true. This to avoid unintented privacy leaks from misconfiguration.

Test suites added for all the new code, all tests pass. Linting run and passes. Docker container with the code has been built and tested successfully in my own Hister installation with user visibility verified for the configured user and non-visibility for an unconfigured user.

# AI DISCLOSURE #

This PR has been made assisted by an LLM: Qwen 3.6 35B A3B in an Opencode+DCP+Superpowers harness locally hosted. All architectual decisions have been made by me (the human Troed writing this), and several iterations to clean up what code to break out into helper functions for better test coverage were made instead of just accepting the LLM's initial output. 

The tests are verbose, but due to the privacy leak risk I wanted to be sure the codepaths have extensive regression tests.

### DIFF
--- a/files/files_test.go
+++ b/files/files_test.go
@@ -1,0 +1,155 @@
+package files
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/asciimoo/hister/config"
+	"github.com/asciimoo/hister/server/model"
+)
+
+func setupTestDB(t *testing.T) *config.Config {
+	t.Helper()
+	cfg := &config.Config{
+		Server: config.Server{
+			Database: "file::memory:",
+		},
+	}
+	err := model.Init(cfg)
+	if err != nil {
+		t.Fatalf("failed to init test DB: %v", err)
+	}
+	return cfg
+}
+
+func TestFindDirUser(t *testing.T) {
+	setupTestDB(t)
+
+	// Create test users
+	u1, err := model.CreateUser("alice", "password123", false)
+	if err != nil {
+		t.Fatalf("failed to create test user: %v", err)
+	}
+
+	dirs := []*config.Directory{
+		{Path: "/home/alice/docs", User: "alice"},
+		{Path: "/home/charlie/docs", User: "charlie"},
+		{Path: "/shared/docs", User: ""},
+		{Path: "/tmp", User: ""},
+	}
+
+	tests := []struct {
+		name    string
+		path    string
+		wantID  uint
+		wantErr bool
+	}{
+		{
+			name:    "file in alice's directory",
+			path:    "/home/alice/docs/file.txt",
+			wantID:  u1.ID,
+			wantErr: false,
+		},
+		{
+			name:    "file in nested alice directory",
+			path:    "/home/alice/docs/subdir/file.txt",
+			wantID:  u1.ID,
+			wantErr: false,
+		},
+		{
+			name:    "file in shared global directory",
+			path:    "/shared/docs/file.txt",
+			wantID:  0,
+			wantErr: false,
+		},
+		{
+			name:    "file in tmp global directory",
+			path:    "/tmp/file.txt",
+			wantID:  0,
+			wantErr: false,
+		},
+		{
+			name:    "file not in any directory",
+			path:    "/other/path/file.txt",
+			wantID:  0,
+			wantErr: false,
+		},
+		{
+			name:    "non-existent user in config",
+			path:    "/home/charlie/docs/file.txt",
+			wantID:  0,
+			wantErr: true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotID, err := FindDirUser(dirs, tt.path)
+			if tt.wantErr {
+				if err == nil {
+					t.Errorf("FindDirUser(%q) expected error, got nil", tt.path)
+				}
+				return
+			}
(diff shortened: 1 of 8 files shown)


files/files_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/asciimoo/hister/config"
+	"github.com/asciimoo/hister/server/model"
+)
+
+func setupTestDB(t *testing.T) *config.Config {
+	t.Helper()
+	tmpDir := t.TempDir()
+	cfg := &config.Config{
+		Server: config.Server{
+			Database: filepath.Join(tmpDir, "test.db"),
```
asciimoo: Perhaps it would be better to use an in-memory db here.

server/indexer/files.go
```diff
@@ -27,13 +28,22 @@ var (
...
+			userID = u.ID
+		}
+		if err := indexDirectory(expanded, dir, userID); err != nil {
 			log.Error().Err(err).Str("directory", expanded).Msg("Failed to index directory")
 		}
 	}
 }
 
-func indexDirectory(dir string, cfg *config.Directory) error {
+func indexDirectory(dir string, cfg *config.Directory, userID uint) error {
```
asciimoo: It would be enough to retrieve the user ID here instead of in `IndexAll`. `IndexAll` does not use that variable besides passing it to `indexDirectory`.

server/indexer/files.go
```diff
@@ -74,7 +84,10 @@ func indexDirectory(dir string, cfg *config.Directory) error {
 	return err
 }
 
-func IndexFile(path string) error {
+func IndexFile(path string, userID ...uint) error {
```
asciimoo: What is the point of making `userID` a variadic parameter? Shouldn't it be simply `uint` instead of `...uint` ?

server/indexer/files_test.go
```diff
@@ -0,0 +1,193 @@
...
+		t.Fatalf("failed to create test file 2: %v", err)
+	}
+
+	// Initialize the indexer with proper data and index dirs
+	idxCfg := &config.Config{
+		App: config.App{
+			Directory: dataDir,
+		},
+		Indexer: config.Indexer{
+			DetectLanguages: false,
```
asciimoo: No need to specify attributes for the `config.Indexer{}` object.

server/indexer/files_test.go
```diff
@@ -0,0 +1,193 @@
...
+			MaxFileSize:     1024 * 1024,
+		},
+		SensitiveContentPatterns: map[string]string{
+			"aws_access_key":      `(^|[\s"'])AKIA[0-9A-Z]{16}([\s"']|$)`,
```
asciimoo: Same here, no need to specify sensitive content patterns. a simple `map[string]string{}` is enough.

hister.go
```diff
@@ -160,10 +177,19 @@ var listenCmd = &cobra.Command{
 			log.Warn().Msg("Using authentication token without https. Token is sent plain-text in network requests.")
 		}
 		if len(cfg.Indexer.Directories) > 0 {
+			// Validate directory user references at startup
+			if err := validateDirectoryUsers(cfg); err != nil {
+				exit(1, err.Error())
```
asciimoo: I would not exit from the application because there are configured directories for non-existent users. Furthermore, no need to do this check at all. It is enough to check the success of `files.FindDirUsers()` below, that function already has user checking built-in.

hister_test.go
```diff
@@ -0,0 +1,95 @@
+package main
```
asciimoo: This file is unnecessary if we remove `validateDirectoryUsers`

asciimoo (changes requested):
Thanks for the contribution. Overall it looks good, but there are some minor issues. Please fix those and run `npm run format` to lint the markdown file.

hister.go
troed: Regarding exit on configured directories for non-existent users I specifically added that so that the admin of the system doesn't believe it's correctly configured to protect the privacy of alice whilst having spellt her username wrong making all her indexed files public instead. Just want to re-check my own view of the severity of that happening :)

asciimoo: >  to protect the privacy of alice whilst having spellt her username wrong making all her indexed files public instead

If the user of the configured directory does not exist, simply throw an error log message and  do not add that directory to the watched directories. It solves all the data leak issues while not blocking a server restart if an admin deletes a user but forgets to remove the related config entry. What do you think?

troed: Yes I had the same thought. The reason for a separate method was to have regression tests on it when building but I agree that this is much cleaner code. I'm making a (local to me, not going to commit) test script that validates all the error scenarios vs log output as well so that I know it's working fully for this PR. Future regression testing on startup conditions is outside the scope here I think.

asciimoo: > Future regression testing on startup conditions is outside the scope here I think.

I agree.

asciimoo (approved):
Nice, thanks!

PR #454, opened 2026-06-03
feat(server): support format=json on /search

Kartalops:
## Summary
- Add `?format=json` and `Accept: application/json` opt-ins to the `/search` endpoint.
- The opt-ins skip the same-host `Origin` check so CLI tools and ad-hoc HTTP consumers can get a JSON response without spoofing `Origin: hister://` (which gets stripped by some reverse proxies).
- The default behaviour for SPA clients is unchanged: a request without the opt-in still goes through the existing `IsSameHost(origin)` gate.
- `format=json` with no text query returns 400 + a JSON error rather than upgrading to a WebSocket search.

## Testing
- `go build ./...` in `server/` — could not run in this environment: the project pins a toolchain newer than the locally available Go (it tried to download `go1.26` and the toolchain server was not reachable). The change is two `if` branches in a single handler; no new imports, no new types, no signature changes.
- Manual QA: start a hister instance, hit `curl 'http://localhost:4433/search?q=foo&format=json'` and observe a `Content-Type: application/json` body. Hit without `format=json` from a different `Origin` to confirm the gate is unchanged.

## Notes
- No new dependencies. No new routes. No new exported symbols.
- The change is additive: every request that worked before still works.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -780,7 +780,13 @@ func serveSearchWebSocket(c *webContext) {
 
 func serveSearch(c *webContext) {
 	origin := c.Request.Header.Get("Origin")
-	if !c.Config.IsSameHost(origin) {
+	// `?format=json` (or `Accept: application/json`) opts the caller into a
+	// pure JSON response and skips the same-host Origin check, so CLI tools
+	// and ad-hoc HTTP clients can hit the search endpoint without spoofing
+	// a hister:// Origin header.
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Thank you for your contribution. LGTM. Strict same origin check is mainly required for of the websocket part of this function, because there is no CORS check in the browsers for websockets. But this PR don't introduce any security concern this regard as I see.

PR #474, opened 2026-06-16
Semantic result count should = total results + semantic hits if this number is larger than total results

ashprice:
Fixes a bug with the semantic results count. 

I *think* this works correctly, but I'm having trouble producing cases that *only* hit samentic search. :)

### DIFF
--- a/server/indexer/indexer.go
+++ b/server/indexer/indexer.go
@@ -1184,7 +1184,7 @@ func Search(cfg *config.Config, q *Query) (*Results, error) {
 			semanticOnlyCnt++
 		}
 	}
-	r.Total = max(r.Total, uint64(len(r.Documents)) + semanticOnlyCnt)
+	r.Total = max(r.Total, uint64(len(r.Documents))+semanticOnlyCnt)
 
 	return r, nil
 }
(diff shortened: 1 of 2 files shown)


server/indexer/indexer.go
```diff
@@ -1176,6 +1176,18 @@ func Search(cfg *config.Config, q *Query) (*Results, error) {
...
 
+	// Bump the total to reflect semantic matches that are not in the
+	// keyword results.
+	semanticOnlyCnt := uint64(0)
+	for _, sh := range r.SemanticHits {
+		if sh.Document != nil {
+			semanticOnlyCnt++
+		}
+	}
+	if docTotal := uint64(len(r.Documents)) + semanticOnlyCnt; docTotal > r.Total {
```
asciimoo: Wouldn't be prettier to write simply `r.Total = max(r.Total, uint64(len(r.Documents)) + semanticOnlyCnt)` ?

ashprice: Yes.

asciimoo (approved):
Nice, thanks!

ashprice:
sorry I forgot to run the linter :face_exhaling:

Issue #445, opened 2026-05-30
yt-dlp in container image.


sohojmanush:
yt-dlp should be included into the docker image, with granular settings into the UI.

mastermint21:
Since this issue isn't closed yet, I would like to point out that the installed yt-dlp version is really old (stable@2025.03.31). I suppose this is due to alpine not having the latest version in their package repositories for alpine:3.21. I guess the better way would be to download the latest yt-dlp binary directly from their GitHub releases page, then place it into /usr/bin.

asciimoo:
Having an up to date yt-dlp is important.

@mastermint21 could you work on this?

asciimoo:
added by @4evy

PR #485, opened 2026-06-22
[enh] install yt-dlp in docker image

YussufSassi:
A small change to the Dockerfile which installs the yt-dlp binary via apk.

This addresses issue #445.

Implementing the configuration UI would be a pretty big undertaking, so I've made the decision to hold off on implementing it for a few reasons:
1. The scope of the configuration UI is not defined.
2. None of the other extractors are configurable via the UI, so all the groundwork (API routes, Svelte Components, etc.) must be laid out.
3. The extractor is already configurable via the config file.

Point 2 & 3 should be worked out beforehand in the issue.

The yt-dlp extractor remains disabled by default, since other environments may not have the binary available.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,7 +52,7 @@ LABEL org.opencontainers.image.title="Hister" \
 
 WORKDIR /hister
 
-RUN adduser -D -u 65532 hister && mkdir -p /hister/data && chown -R 65532:65532 /hister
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Nice, thanks!

asciimoo:
> Implementing the configuration UI would be a pretty big undertaking,

Yes, I agree. Probably it would be good to provide a more general UI for the whole config file on the long run.

Thank you for your contribution!

PR #492, opened 2026-06-24
Add public mode

### LINKED ISSUE #472: Make it possible to have a public instance of Hister
I understand Hister is focused on making a great *personal* web search engine.

I'd like to use it for something a bit different: a community search engine. Thanks to its nice indexing capabilities, it would be easy to make a search engine which is optimized on a specific set of pages.
To make this possible, I should be able to make it impossible to change the database from the public web interface.

Would that be possible?

asciimoo:
Introduce public mode where all the users have read-only access to the web interface, but only authenticated users can modify the content.

Test it by adding `public = true` to the config's `app` section.

Closes #472

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -51,6 +51,7 @@ type App struct {
 	SearchURL              string `yaml:"search_url" mapstructure:"search_url"`
 	AccessToken            string `yaml:"access_token" mapstructure:"access_token"`
 	UserHandling           bool   `yaml:"user_handling" mapstructure:"user_handling"`
+	Public                 bool   `yaml:"public" mapstructure:"public"`
 	LogLevel               string `yaml:"log_level" mapstructure:"log_level"`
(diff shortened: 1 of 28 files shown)


YussufSassi:
Works great from my quick testing, I just found three frontend changes that would make sense I think:

1. The Authentication Token should not be stored in `localStorage`, an HttpOnly cookie would better. (`localStorage` is susceptible to XSS attacks) 
2. The "Save Token" and Login/Logout buttons should have a `cursor: pointer` for better UI cohesion.
3. There should be some more info (like toasts or flash messages) for some cases (like invalid auth token, logout). Redirecting without any info could be a bit confusing for the user.

asciimoo:
Thanks for the review, all points are valid. Fixing soon.

YussufSassi:
LGTM 👍🏻

asciimoo:
Great, thanks again for the review. I'm merging it and we can refine it later on.

PR #494, opened 2026-06-25
[enh] implement generic stackexchange question/answer extractor

YussufSassi:
This PR addresses part of issue #305 which asked for a generic stack exchange question & answer extractor.

It works based on matching the hostname to a list of seven apex domains (including their subdomains) used by the stackexchange network. I think this pattern matching is robust enough to work reliably, I initially had added some DOM heuristics as well, but figured it might be too prone to false positives and removed it.

I've also gone ahead and replaced the specific stackoverflow extractor, as this extractor deprecates it. A migration for existing users of that extractor shouldn't be necessary, and the transition should be invisible to them. 

Here's a screenshot of what an indexed stackexchange url looks like in Hister:
image

> [!IMPORTANT]
> Using `hister index <stack-exchange-url>` will not work without the `--no-robots` flag  due to their robots.txt file. However indexing via the browser extension should work by default.


asciimoo (approved):
This is fantastic, thank you!

Later on we could further enhance it by adding metadata attributes (question/answer timestamps, tags, perhaps author)

YussufSassi:
The metadata attributes should be available already, author and date added at least.

YussufSassi:
Oh and the docs page @ https://hister.org/docs/extractors must be updated as well, I forgot to do this in my commit.

asciimoo:
>  The metadata attributes should be available already, author and date added at least.

Currently it is only available for the question in the preview header, but I think date is important in the answer as well to check how old is the answer.

> Oh and the docs page @ https://hister.org/docs/extractors must be updated as well, I forgot to do this in my commit.

No worries, I'll do it.

YussufSassi:
oh yeah, those changes make a lot of sense, PR coming asap!

PR #495, opened 2026-06-25
Improve GitHub extractor

RensOliemans:
From #305 , I started with a stab at improving the GitHub extractor. The PR is still very much WIP, and I thought I'd create it now because I wanted to ask this question: do you want me to split `github.go` into separate files? Like `issue.go` for parsing the `/owner/repo/issues/:id` url, etc.

I actually have two other questions about the existing code (I'm quite new to Go):
1. What is the purpose of `Preview`? Should this be unique for each type of URL as well?
2. What was the purpose of the `b string.Builder`? Should I keep it, or use only metadata?


asciimoo:
Nice!

> do you want me to split github.go into separate files?

It's up to you. Having long files is idiomatic in Go if there is no real logical separation that would make sense (e.g. our `server.go`). Perhaps in this case it makes sense to put different handlers to different files, but it isn't necessary.

> What is the purpose of Preview? Should this be unique for each type of URL as well?

`Preview()` produces the result preview displayed in the right pane of the results view.

> What was the purpose of the b string.Builder? Should I keep it, or use only metadata?

It collects important info from the page that is assigned to the `document.Text` field. The advantage of assigning the content to `document.Text` is that it instantly becomes searchable, while the metadata fields are not searchable by default if you don't use query filtering for those fields (e.g. `metadata.language:go` search query).

So information that should be instantly searchable should go to `document.Text` (e.g. project description) and data attributes that just provide additional information in the preview (or special filtering) should go to the metadata (e.g. stars).

asciimoo:
Let me know if I can start the review.

RensOliemans:
Thanks for the info. I'll let you know when I'm happy with it!

RensOliemans:
Should I think about changing GitHub page structure? For PRs especially the html github uses is quite ugly. I use the selector `.js-comment-container` to find the body and comments and `[data-status]` to obtain the state of a PR. However I can imagine this changing in the future. (How) should I deal with this?

asciimoo:
> I use the selector `.js-comment-container` to find the body and comments and `[data-status]` to obtain the state of a PR. 

These selectors look fine. Rule of thumb is to find selectors as generic as possible. This way we can reduce the potential break when something changes, but not much else we can do.

> However I can imagine this changing in the future.

Sure, this is natural when scraping.

>  (How) should I deal with this?

No need to put any extra effort to handling changes for now. We don't know what changes are coming, so it's hard to prepare. My usual strategy is to wait until it breaks and adjust if we see the changes.

RensOliemans:
OK I'm ready for (a first round of) feedback! I have skipped Preview for now since the default is acceptable I think and it would make it a bit more complicated. The new types of pages I support are seen in `github_test.go`:
- https://github.com/asciimoo/hister/issues
- https://github.com/asciimoo/hister/issues/:id
- https://github.com/asciimoo/hister/pull/:id

Please don't hesitate with feedback, I'm new to Go and I like to learn!

asciimoo:
It looks clean. There are minor formatting issues reported by the "PR Lint" github workflow. And as I see the regexps don't match some valid URLs like:
```
https://github.com/asciimoo/hister/
https://github.com/asciimoo/hister?tab=readme-ov-file
```

RensOliemans:
Thanks. I relaxed the regexps a bit. They can always be extracted into multiple regexps per url if necessary, if they get too unwieldy - but I think they're okay for now, wdyt?

I fixed the linting issues.

asciimoo:
Great work, thank you!

I hope we can receive more contributions from you. <3

asciimoo (approved):

asciimoo:
Not sure why, but the preview of this github issue isn't the same as the others: https://github.com/asciimoo/hister/issues/524 .
In comparison the preview of e.g. https://github.com/asciimoo/hister/issues/500 is beautiful.

RensOliemans:
Thanks for merging! And the issue you describe is interesting, I have not yet
figured that one out...

I haven't touched the default Preview because of two reasons:
1. the default was very often beautiful (see 500)
2. this made it difficult to override the Preview and not regress in any way.

However that doesn't explain why 524 is so different from all the rest. You
can also compare it to the very similar
https://github.com/asciimoo/hister/issues/505, also a closed issue with a
relatively simple description, but even that one has a very different preview.

I'll look into this a bit more!

asciimoo:
Yes, I've checked a few issues and most of them are completely fine, but a few are rendered similarly wrong as issue 524. I tried to find some kind of connection between the good/bad cases, but no luck so far.

RensOliemans:
Ah, found it. Since I didn't modify Preview, it only works on repositories, by virtue of this in Preview:
https://github.com/asciimoo/hister/blob/81753a5ba459ed95e065e44a95ae69c00c11aabc/server/extractor/extractors/github/github.go#L155-L158

Therefore, `extractor.go:Preview` skips over this extractor and it defaults to the `readabilityExtractor`. Relevant code:
https://github.com/asciimoo/hister/blob/81753a5ba459ed95e065e44a95ae69c00c11aabc/server/extractor/extractor.go#L223-L242

And `readabilityExtractor` uses [go-readability](https://codeberg.org/readeck/go-readability) (I think), and the difference is in how `go-readability` works:
```bash
rens@nixos ~/g/bin> ./go-readability -t https://github.com/asciimoo/hister/issues/524
failed to detect readable content on the page
rens@nixos ~/g/bin [1]> ./go-readability -t https://github.com/asciimoo/hister/issues/505
On firefox extension version 0.26.1, when turning on "show indicator for indexed pages" and start a manual indexing by clicking "index this page now" on any webpage, the extension will report an error Reindex failed: Incorrect argument types for action.setBadgeTextColor., and the indexing will fail, with server/server.go:495 > WEB LoadTimeMS=56.275201 Method=POST Status=406 URL=/api/add on the server. However, when the indicator is turned off, the indexing will succeed.
I'm not sure if this is specific to firefox so I'll point it out explicitly. Running on Firefox Developer Edition 153.0b6.
```

What would you recommend? I started on Preview, but I refrained because in the majority of cases the current behavior is great, and when github.go:Preview returns something it will never go to the current behavior. So github.go:Preview better be *good*. Or is there another way that could combine the two?

RensOliemans:
Ok when thinking a bit more about it, it's actually not too difficult to get Preview to work nicely with Issues. I was confused since I thought that it would override the current default (readabilityExtractor) behavior in the other cases (Pull Requests, Wiki pages, etc), but of course it won't as long as we only create a custom Preview for Issue pages.

I can work on this. Perhaps this weekend but perhaps it'll take a bit longer.

asciimoo:
> Therefore, extractor.go:Preview skips over this extractor and it defaults to the readabilityExtractor

I'm still failing to understand how could this result in displaying most of the issues well, but some of them wrong.

asciimoo:
> I can work on this. Perhaps this weekend but perhaps it'll take a bit longer.

Thanks, but no need to rush, the current version of the GitHub extractor is already miles better than before. I didn't mention it because this is a serious issue, I've just noticed that there is an inconsistent behavior.
I appreciate if you take a look at it if you have some free time, but no pressure. =]

RensOliemans:
> > Therefore, extractor.go:Preview skips over this extractor and it defaults to the readabilityExtractor
> 
> I'm still failing to understand how could this result in displaying most of the issues well, but some of them wrong.

Perhaps I misunderstand you, but isn't that because of the behaviour of `go-readability` which is used by readabilityExtractor? See these two cases:

```bash
$ ./go-readability -t https://github.com/asciimoo/hister/issues/524
failed to detect readable content on the page

$ ./go-readability -t https://github.com/asciimoo/hister/issues/505
On firefox extension version 0.26.1, when turning on "show indicator for indexed pages" and start a manual indexing by clicking "index this page now" on any webpage, the extension will report an error Reindex failed: Incorrect argument types for action.setBadgeTextColor., and the indexing will fail, with server/server.go:495 > WEB LoadTimeMS=56.275201 Method=POST Status=406 URL=/api/add on the server. However, when the indicator is turned off, the indexing will succeed.
I'm not sure if this is specific to firefox so I'll point it out explicitly. Running on Firefox Developer Edition 153.0b6.
```

asciimoo:
Yes, the root cause is the readability plugin. What I don't understand is why is it differentiate the two pages. Anyway if we go to the direction of adding a custom `Preview()` to the GitHub extractor, it doesn't really matter - but it is still odd.

RensOliemans:
Ah, I was misunderstanding you then. A quick glance suggests that the plugin assigns a "readability score" to each page / elements within the page, with the goal of showing the readable text, seen here: https://codeberg.org/readeck/go-readability/src/commit/29522a0e224f27b5ce43c65b7aadf60ad7d04e5d/parser.go#L953

The text in #524 does not get assigned a high enough score to be considered readable. Perhaps because a large part of the text is a link (high linkDensity), has no comma's and little text in general, or some other magic.

This is corroborated by passing `-f`. From `go-readability --help`:
```
Flags:
  -f, --force         continue parsing documents that failed the readerable check
```



```
rens@nixos ~/g/bin> ./go-readability -t -f https://github.com/asciimoo/hister/issues/524
warning: the page might not have article contents
I noticed on the community-script website that Hister can be installed quite easily on LXC

https://community-scripts.org/scripts/hister

bash -c "$(curl -fsSL https://raw.githubusercontent.com/community-scripts/ProxmoxVED/main/ct/hister.sh)"
```

Perhaps hister could do the same?

asciimoo:
Oh, that's good to know, thank you for the investigation, it explains the behavior. I have to do further tests to make sure that the output of the `--force` version is better than falling back to the basic extractor.

PR #496, opened 2026-06-25
add org mode extractor

cashmeredev:
This PR adds a new file extractor for .org files. I tried to stay as close as possible to the existing Markdown extractor. It's worth mentioning that this PR introduces [go-org](https://github.com/niklasfasching/go-org) as a dependency to the project. go-org is also used by [Hugo](https://github.com/gohugoio/hugo/blob/master/markup/org/convert.go) to render Org files to HTML. I tested it against my personal wiki, and all files render without issues.

@asciimoo already mentioned adding an interface to handle local file types better. This is still missing. Let me know how we could introduce it, and I will also include it in this PR.

image

Kind regards,
cashmere

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -104,6 +104,7 @@ require (
 	github.com/mschoch/smat v0.2.0 // indirect
 	github.com/muesli/ansi v0.0.0-20230316100256-276c6243b2f6 // indirect
 	github.com/muesli/cancelreader v0.2.2 // indirect
+	github.com/niklasfasching/go-org v1.9.1 // indirect
 	github.com/pelletier/go-toml/v2 v2.4.0 // indirect
 	github.com/pierrec/lz4/v4 v4.1.27 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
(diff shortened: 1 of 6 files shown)


asciimoo:
Great, thanks! The PR is clean. Please fix the go lint errors and it is good to go.

> @asciimoo already mentioned adding an interface to handle local file types better. This is still missing. Let me know how we could introduce it, and I will also include it in this PR.

I think it is out of the scope of this PR. Basically my idea is to create a `server/indexer/filetypes` package at first where we can handle all the different file type parsers. The package should only have a public `CreateDocument(doc *Document, content []byte) (*Document, error)` (or `ParseDocument` - not sure about the naming) method. What do you think?

cashmeredev:
> Great, thanks! The PR is clean. Please fix the go lint errors and it is good to go.
> 
> > @asciimoo already mentioned adding an interface to handle local file types better. This is still missing. Let me know how we could introduce it, and I will also include it in this PR.
> 
> I think it is out of the scope of this PR. Basically my idea is to create a `server/indexer/filetypes` package at first where we can handle all the different file type parsers. The package should only have a public `CreateDocument(doc *Document, content []byte) (*Document, error)` (or `ParseDocument` - not sure about the naming) method. What do you think?

We could open a seperate PR for that, I don't mind! :)

asciimoo (approved):
Awesome, thanks!

PR #497, opened 2026-06-26
[fix] add metadata to stackexchange answers

YussufSassi:
as mentioned in #494, this PR adds metadata (date, author, votes/score) to stack exchange answers.

image

### DIFF
--- a/server/extractor/extractors/stackexchange/stackexchange.go
+++ b/server/extractor/extractors/stackexchange/stackexchange.go
@@ -3,6 +3,7 @@ package stackexchange
 
 import (
 	"fmt"
+	stdhtml "html"
 	"net/url"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


asciimoo:
Great, thanks! Could you please add the upvote count as well? It is often useful to check if a newer answer has more upvotes than the accepted one, espcially in HTML/CSS/JS questions where the question is often old and there are newer and better solutions to a specific problem.

YussufSassi:
The votes/score number is already basically the upvote amount.

It's this number between the up/downvote buttons: 
image

It allows users to check which answer had the most upvotes.

asciimoo:
Oh, sorry. I've missed that you parse for the answers in `postMetaLine()`.

asciimoo (approved):
Thanks!

PR #509, opened 2026-07-04
[enh] add rfc dataset

0x00cl:
Adds the RFC dataset.

It includes RFC from 1 to 10014 (latest one). It does not include drafts or work in progress.

root RFCs: 9787
inline-errata: 1728
Total: 11515

For what its worth, its about 1.53 GiB and importing it in my machine in a clean instance:

```
$ time ./hister import ~/Downloads/rfcs.json 
✓ Imported 11515 document(s)

real	31m52.181s
user	0m14.052s
sys	    0m1.584s
```

### DIFF
--- a/webui/website/src/content/docs/browser-extension.md
+++ b/webui/website/src/content/docs/browser-extension.md
@@ -58,6 +58,7 @@ Clicking the extension icon opens the popup, which provides quick access to the 
 | **Authenticate Extension** button            | Copy session cookies from the logged-in Hister web UI to authenticate the extension (user handling only).             |
 | **Settings icon** (⚙)                        | Expand an inline form to view and update the Server URL without opening the full options page.                        |
 | **Show indicator for indexed pages** setting | Show a `✓` badge on pages that are already indexed. This is available after opening the settings view from the popup. |
+| **Submit as public documents** setting       | Store newly indexed pages as public documents with user id `0`. This appears after the extension is authenticated.    |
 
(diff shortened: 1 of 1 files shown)


asciimoo:
Hmm, this import time is far from ideal, I should try to improve it.

Do you mind if I publish the dataset 7zipped on hister.org?

0x00cl:
No problem.

FYI I used https://github.com/0x00cl/rfc-dataset script to generate the `rfcs.json` using the `rsync`ed data.

asciimoo:
Thanks! Hister already has a `scripts/fetch_dataset.sh`. The best would be to integrate your script to our dataset fetcher (I'm thinking about creating something like a `scripts/fetch_dataset_helpers` folder and adding your python script there and calling it from `fetch_dataset.sh`). Are you interested in implementing it?

0x00cl:
**I'll try to give it go and see if I can come up with something that makes sense**, because as it is right now it uses `rsync` to download the RFCs, which is probably not available on Windows, and would require either Cygwin or MSYSS2.

Also, personally, I'd rather not add another language requirement such as `Python` to the project, which again Windows will have problems if its not installed or `uv` installed.

asciimoo:
@0x00cl There is no windows support requirement. These are just helper scripts to make dataset creation reproducible - scripts are not part of the production codebase. It is totally fine to leave the python script as it is, no need to introduce extra complexity.

0x00cl:
I might have closed it by mistake..., I'll make a new PR that includes the script to fetch RFCs.

PR #513, opened 2026-07-05
[enh] add rfc fetch dataset command

0x00cl:
I added the `fetch_rfcs()` function to the `fetch_dataset.sh` script.

It uses rsync, as [suggested by rfc-editor](https://www.rfc-editor.org/series/rfc-download/), to download the RFCs, html, text and json (metadata).

Then these files are processed using Python which I created the directory `fetch_dataset_helpers` and this Python scripts ends up generating a json file `rfcs.json` which can be imported into hister.

When running the script, the `rfcs.json` ended up being 1647025877 bytes (1.53 GiB) with a total of 11515 documents, 9787 being RFCs and 1728 inline-errata.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,5 @@ tmp/
 
 .test/
 .pre-commit-config.yaml
+scripts/fetch_dataset_helpers/my-rfc-mirror/
+scripts/**/__pycache__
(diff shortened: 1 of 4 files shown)


asciimoo:
Fantastic, thanks a lot! Could you include the original dataset definition JSON as well, which was part of the first PR?

0x00cl:
Sure, I added the original .json definition.

asciimoo (approved):
Thank you, LGTM.

Issue #459, opened 2026-06-05
Example systemd unit file for non-nix systems?


OmenWild:
It would be helpful to have an example systemd unit file for non-nix systems too. I see the hardening options in the nix/nixos.nix file, and I could _probably_ recreate them, but it would be useful if there was a full sample unit file that had all the same hardening options.

Thank you,
   Omen

mascalzeeno:
Hi, I have made one which works on my machine. I copied what is present in the nix file, apart from the following:
- EnvironmentFile is absent
- I didn't know what this line does so I ignored it: `environment = histerEnv config.services.hister;`
- AmbientCapabilities and CapabilityBoundingSet are empty as I have it listen to port 4433
- the user and group are hardcoded to the ones I'm using

```
[Unit]
Description="Hister web history service"

[Service]
Type=simple
User=www-data
Group=www-data
ReadWritePaths=/var/www/hister
ExecStart=/var/www/hister/hister listen
Restart=on-failure
AmbientCapabilities=
CapabilityBoundingSet=
NoNewPrivileges=true
ProtectSystem=strict
PrivateTmp=true
PrivateDevices=true
ProtectKernelTunables=true
ProtectKernelModules=true
ProtectControlGroups=true
ProtectClock=true
ProtectHostname=true
ProtectProc=invisible
ProcSubset=pid
LockPersonality=true
RestrictNamespaces=true
RestrictRealtime=true
RestrictSUIDSGID=true
RestrictAddressFamilies=AF_INET AF_INET6 AF_UNIX
SystemCallArchitectures=native
SystemCallFilter=@system-service ~@privileged
MemoryDenyWriteExecute=true
UMask=0077

[Install]
WantedBy=multi-user.target
```

OmenWild:
This was a big help. Thank you.

asciimoo:
Added in #515.

PR #515, opened 2026-07-05
Add a sample systemd unit

ahfoysal:
Adds a sample systemd unit for non-Nix Linux installs, matching the hardening options already used by the NixOS module as closely as a reusable example can.

This gives #459 a copyable starting point without changing the packaged service.

Checked:
- npm exec prettier -- --check webui/website/src/content/docs/installing.md
- npm -w @hister/website run build

I could not run systemd-analyze on this macOS host.

AI disclosure: I used Codex to help prepare this small docs/service-file patch and compare it against the existing NixOS unit settings.

### DIFF
--- a/contrib/systemd/hister.service
+++ b/contrib/systemd/hister.service
@@ -1,0 +1,43 @@
+[Unit]
+Description=Hister web history service
+After=network.target
+
+[Service]
+Type=simple
+User=hister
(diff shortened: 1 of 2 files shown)


asciimoo:
Nice! Thanks for the contribution and for being transparent about how you use AI.

PR #517, opened 2026-07-06
[fix] install yt-dlp from upstream releases - #445

4evy:


Dockerfile
```diff
@@ -52,7 +71,8 @@ LABEL org.opencontainers.image.title="Hister" \
 
 WORKDIR /hister
 
-RUN apk add --no-cache yt-dlp && adduser -D -u 65532 hister && mkdir -p /hister/data && chown -R 65532:65532 /hister
+RUN adduser -D -u 65532 hister && mkdir -p /hister/data && chown -R 65532:65532 /hister
+COPY --from=ytdlp /usr/local/bin/yt-dlp /usr/local/bin/yt-dlp
```
asciimoo, line 75: What is the purpose of this line? Seems like the source and the destination parameters are the same.

4evy: It copies from the ytdlp stage into the final image. Same path, but different stage/filesystem

asciimoo (approved):
Great, thanks!

Issue #453, opened 2026-06-03
Add URL-Only Collection Mode to Hister.


sohojmanush:
I like to have a feature where Hister only stores URL (Browsing History) without indexing the page. A feature like that will be highly appreciated.

PR #528, opened 2026-07-10
[enh] Processing file checks field already set in JSON

0x00cl:
When processing the JSONL, instead of extracting the text again and trying to detect the language again, it first checks whether the document has a text set and if it has a language set. Also instead of rewriting the time it was indexed, it will check whether this value is 0 or not. For example if a user exports to back up their data, and later decide to import it, it will keep the time it was originally indexed, while datasets or other type of JSONL created should not have an added time key, so it actually sets the current time.

I used [hyperfine](https://github.com/sharkdp/hyperfine) to benchmark these changes with the following command `hyperfine -w 0 -r 5 -p "sync; echo 3 | sudo tee /proc/sys/vm/drop_caches; rm -rf ~/.config/hister && ./hister -l error listen & sleep 2" -C "pkill hister || true && rm -rf ~/.config/hister" --parameter-list file datasets/go-stdlib.json,datasets/python-docs.json,datasets/mdn-html.json,datasets/nodejs-api.json --export-markdown benchmark-results.md "./hister import {file}"`

* No warmup
* 5 runs per dataset import
* Always starts from a clean state (deletes hister storage)
* Clears  OS cache

Before:

| Command | Mean [s] | Min [s] | Max [s] |
|:---|---:|---:|---:|
| `./hister import datasets/go-stdlib.json` | 21.741 ± 7.376 | 15.799 | 33.927 |
| `./hister import datasets/python-docs.json` | 27.931 ± 0.474 | 27.435 | 28.697 |
| `./hister import datasets/mdn-html.json` | 22.548 ± 1.360 | 20.593 | 23.841 |
| `./hister import datasets/nodejs-api.json` | 14.762 ± 2.722 | 12.715 | 18.583 |

This PR:

| Command | Mean [s] | Min [s] | Max [s] |
|:---|---:|---:|---:|
| `./hister import datasets/go-stdlib.json` | 18.752 ± 9.692 | 12.767 | 35.990 |
| `./hister import datasets/python-docs.json` | 23.677 ± 2.236 | 20.372 | 26.624 |
| `./hister import datasets/mdn-html.json` | 19.385 ± 4.954 | 15.705 | 27.549 |
| `./hister import datasets/nodejs-api.json` | 11.841 ± 3.623 | 9.091 | 17.308 |

I used GLM-5.2 to make the changes and it decided to add tests for `document.go`. I did review the changes.

### DIFF
--- a/server/document/document.go
+++ b/server/document/document.go
@@ -138,10 +138,12 @@ func (d *Document) Process(ld LanguageDetector, extractFn func(*Document) error)
 		return errors.New("invalid URL: missing scheme/host")
 	}
 	d.normalizeWebURL(pu)
-	d.Added = time.Now().Unix()
+	if d.Added == 0 {
+		d.Added = time.Now().Unix()
+	}
 	d.Type = types.Web
 	d.Domain = pu.Hostname()
-	if d.HTML != "" {
+	if d.HTML != "" && d.Text == "" {
 		if err := extractFn(d); err != nil {
 			return err
 		}
@@ -207,7 +209,9 @@ func (d *Document) processFile(ld LanguageDetector) error {
 
 // finalizeDocument sets the document language and marks it as processed.
 func (d *Document) finalizeDocument(ld LanguageDetector) {
-	d.Language = ld.DetectLanguage(d.Text)
+	if d.Language == "" || d.Language == UnknownLanguage {
+		d.Language = ld.DetectLanguage(d.Text)
+	}
 	d.processed = true
 }
 
--- a/server/document/document_test.go
+++ b/server/document/document_test.go
@@ -1,0 +1,168 @@
+package document
+
+import (
+	"errors"
+	"testing"
+)
+
+// fakeDetector records the text it was asked to detect and returns a fixed
+// language, so tests can assert whether detection ran.
+type fakeDetector struct {
+	called   bool
+	gotText  string
+	language string
+}
+
+func (f *fakeDetector) DetectLanguage(s string) string {
+	f.called = true
+	f.gotText = s
+	return f.language
+}
+
+const sampleHTML = `<html><head><title>Raw Title</title></head><body><p>raw body</p></body></html>`
+
+func TestProcessSkipsExtractionWhenTextPresent(t *testing.T) {
+	d := &Document{
+		URL:   "https://example.com/a",
+		HTML:  sampleHTML,
+		Text:  "pre-extracted text",
+		Title: "Imported Title",
+	}
+	ld := &fakeDetector{language: "fr"}
+
+	extractCalled := false
+	extractFn := func(*Document) error {
+		extractCalled = true
+		return nil
+	}
+	if err := d.Process(ld, extractFn); err != nil {
+		t.Fatalf("Process() unexpected error: %v", err)
+	}
+	if extractCalled {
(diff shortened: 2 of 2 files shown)


server/document/document.go
```diff
@@ -138,10 +138,12 @@ func (d *Document) Process(ld LanguageDetector, extractFn func(*Document) error)
 		return errors.New("invalid URL: missing scheme/host")
 	}
 	d.normalizeWebURL(pu)
-	d.Added = time.Now().Unix()
+	if d.Added == 0 {
```
asciimoo, line 141: I agree, this is the proper way to handle `Added`, but currently this field functions more like a last update indicator. With your change we cannot identify if a page got updated. I'd keep changing the `Added` behavior to a separate pull request where we introduce an `Updated` field as well, which can be used as how we use `Added` currently, so the history and result ordering can remain the same.

asciimoo:
Instead of checking if the language and text field has value, I'd introduce a `Processed` `bool` attribute to the `Document` object. This should not be stored in the index (noindex, nostore attributes in the index.go mapping definition), but can be used by the `export` command to indicate that the document requires no further processing. The import could check this property and skip language detection and extractor execution.

The advantage of this approach is that it still allows to reprocess documents when those contain changes or when an extractor got updated or a `reindex` action is performed.

What do you think?

server/document/document.go
0x00cl: Ok, makes sense, I'll change it.

0x00cl:
Oh I didn't realize `Process` was used to index URLs too and not only the `import` command.

Doesn't `Document` already have one called `processed`?

https://github.com/asciimoo/hister/blob/afee231342989f92c1e8aad23969bffb1c71974f/server/document/document.go#L47

As it is right now, without this PR, default is "forcing" every document imported to be processed and ignores other key/values from the JSONL file and there is no way to opt out of this.

So, adding a way to tell the server to not process the current file would be a good idea and simply store as is, the URL should be enough and maybe the title (I believe a URL only mode was requested in https://github.com/asciimoo/hister/issues/453).

Opting out of processing the URL being indexed might allow users to store links only and also allow users who exported their data and want to import their data, opt out of their data being reprocessed and speed up importing.

So, I'd agree on adding an attribute to the document (or re-using one) that can tell the server to not process the current document. If you go this way I believe this PR shouldn't be merged and should be closed as a document will either need to be reprocessed completely or not and not a field basis.

asciimoo:
> Oh I didn't realize Process was used to index URLs too and not only the import command.

What do you mean?

> Doesn't Document already have one called processed?

It has but it is private, it's not possible to specify it from outside of the module. Making it public would allow us to set it in the `import` command.


> Opting out of processing the URL being indexed might allow users to store links only and also allow users who exported their data and want to import their data, opt out of their data being reprocessed and speed up importing.

It think it is already supported by the backend if you don't send any `Text` or `HTML` content.

> If you go this way I believe this PR shouldn't be merged and should be closed as a document will either need to be reprocessed completely or not and not a field basis.

Alright, I can implement this approach if you don't want to rewrite this PR to implement the discussed solution.

asciimoo:
@0x00cl I've made `Processed` global and modified `export` to use it. Could you run a new benchmark with the batch and `Processed` changes and compare it to the original version? Or, could you share your benchmark script?

PR #532, opened 2026-07-11
feat(ui): support bulk rule entry

Codycody31:
Adds a bulk rule entry dialog and button to the indexing rules section.

image
image


asciimoo (approved):
Useful feature, thanks.

It would be good to refactor the rule handling in the future and create dedicated, typed class(es) for rules that can handle rule related operations in one place.

PR #533, opened 2026-07-12
fix: keep search input visible

Codycody31:
Just fixes the search input box getting too small/thin depending on the screen ratio/size

Before: image
After: image


asciimoo (approved):
Nice, thanks!

PR #559, opened 2026-07-27
Add option to auto-index page only once

g2mt:
This PR adds a checkbox to the web extension to skip auto indexing if the page is already indexed.

image

Right now the cjsMsgHandler function nests a lot of callbacks. For clarity I refactored the relevant block into an anonymous async function. I would rather the handler be one async function but that's outside the scope of this PR.

Motivation: I use hister and the web extension primarily as an archival tool. There are a lot of sites (i.e. link aggregator sites) that I don't want archived. In contrast long form content like articles or blog posts are read once and usually do not receive updates. Adding this option allows me to skip the first set of sites while archiving the latter.

AI Disclosure: pi+deepseek was used to modify the UI and add the input checkbox. The rest was coded by hand.

### DIFF
--- a/webui/ext/src/background/background.ts
+++ b/webui/ext/src/background/background.ts
@@ -544,7 +544,7 @@ function cjsMsgHandler(request, sender, sendResponse) {
           let r;
           try {
             r = await sendPageData(u + 'api/add', pageData, customHeaders);
-          } catch(err) {
+          } catch (err) {
             setErrorBadge(sender.tab.id);
             sendResponse({ error: err.message });
             return;
--- a/webui/ext/src/background/background.ts
+++ b/webui/ext/src/background/background.ts
@@ -448,6 +448,7 @@ function cjsMsgHandler(request, sender, sendResponse) {
       'histerURL',
       'histerToken',
       'indexingEnabled',
+      'indexOnlyOnce',
       'histerCustomHeaders',
       'showIndexedBadge',
       'submitPublicDocuments',
@@ -455,6 +456,7 @@ function cjsMsgHandler(request, sender, sendResponse) {
     .then((data) => {
       let u = data['histerURL'] || '';
       const indexingEnabled = data['indexingEnabled'] !== false;
+      const indexOnlyOnce = data['indexOnlyOnce'] !== false;
       const showIndexedBadge = data['showIndexedBadge'] === true;
       const customHeaders = getCustomHeaders(data);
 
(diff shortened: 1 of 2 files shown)


asciimoo:
How does the proposed change allow skipping link aggregator sites?

Hister has skip rules exactly for this purpose.

g2mt:
> How does the proposed change allow skipping link aggregator sites?
> 
> Hister has skip rules exactly for this purpose.

I messed up my explanation. The primary goal is to have articles, etc. be archived only once on first visit. By default, hister seems to update the archived version every time it is navigate to. It's not meant to be a replacement for the skip rules.

asciimoo:
Could you please move it to the settings tab and use the same toggle ui component as the indexed page indicator?

g2mt:
Done.

asciimoo:
After a more thorough review, I'm not entirely sure if this is the right approach.

The current solution numerous flaws:
 - If I understand correctly, the option is enabled while the UI says it is disabled (`data['indexOnlyOnce'] !== false;` is going to be true for existing installations where the variable isn't set)
 - When the HEAD request finds a document, the async function returns without calling sendResponse. The listener has returned true, so Chrome keeps the message channel open expecting a response.
 - The existence check can inspect the wrong owner (`HEAD /api/document` uses `GetByURLAndUser`, which falls back from the authenticated user to the global document)
 - The extension checks the original URL. The server strips UTM parameters and normalizes the resulting query string before storage. This can lead to false negative reports
 - PDFs ignore the setting

Probably a better approach would be to check duplicated submissions server side instead.

There could be a config settings to save only new documents. This avoids the extra HEAD request, URL normalization mismatch, ownership confusion, and race condition. It also guarantees a response to the content script.

What do you think?

g2mt:
> There could be a config settings to save only new documents. This avoids the extra HEAD request, URL normalization mismatch, ownership confusion, and race condition. It also guarantees a response to the content script.

Sure. Are you suggesting a global configuration field? I think ignoreOnlyOnce should be saved to the pageData before passing to api/add, since this is an extension-specific setting.

asciimoo:
> Are you suggesting a global configuration field? I think ignoreOnlyOnce should be saved to the pageData before passing to api/add, since this is an extension-specific setting.

This is what I'm not sure about. Perhaps you are right in multi-user environments. It can happen that user A wants standard indexing and user B wants to index only once.

BTW, what is exactly the advantage of rejecting documents that are already in the index? What is your use-case where it is preferred over the current indexing behavior?

g2mt:
> BTW, what is exactly the advantage of rejecting documents that are already in the index? What is your use-case where it is preferred over the current indexing behavior?

For archival purposes. With this, if the online version of a page is deleted and returns a 404, then I could still rely on an offline copy without it being modified. Also it's kind of an optimization thing. There are pages or reference documents which I frequently view, and I don't need these documents to be processed on the server again.

asciimoo:
> if the online version of a page is deleted and returns a 404, then I could still rely on an offline copy without it being modified.

The extension does not submit non 200 status code pages if it can detect it (the detection isn't reliable on Firefox).

>  Also it's kind of an optimization thing. There are pages or reference documents which I frequently view, and I don't need these documents to be processed on the server again.

There is no write operation if the submitted document is the same as the stored one and with the new approach we would still submit the document to the backend, so it isn't really optimizes anything.

g2mt:
I have realized my patch didnt make sense for a multi user environment and various other reasons. I'm still using it as a personal fork, but I have closed the PR. sorry for wasting your time!

PR #561, opened 2026-07-29
Add required scope param to Google OAuth redirect

jrocketfingers:
[Google OAuth 2.0 endpoint](https://developers.google.com/identity/protocols/oauth2/web-server#httprest_1:~:text=web%20server%20applications.-,scope,Required,-A%20space%2Ddelimited) expects scope param as **required**. This commit adds it using the pre-existing implementation of GetScope, making it in-line with GitHub OAuth implementation, but diverges from OIDC implementation which uses scopes provided from the request itself.

Without this param, the OAuth flow fails with [`invalid_request`](https://developers.google.com/identity/protocols/oauth2/web-server#:~:text=The%20request%20was%20missing%20required%20parameters).

Few review notes:
- I don't really write go, so review with heavy scrutiny, I'm happy to learn.
- I couldn't find a consistent approach, between GitHub, OIDC and Google auth, I'm looking for input.

Tests passed, and I'm currently running this code on my instance, so I thought I'd sned the changes back

### DIFF
--- a/server/oauth/google.go
+++ b/server/oauth/google.go
@@ -31,6 +31,13 @@ func (g GoogleOAuth) Prepare(_ context.Context, _ *PrepareRequest) error { retur
 func (g GoogleOAuth) GetRedirectURL(req *RedirectURIRequest) string {
 	params := &url.Values{}
 
+	scopeName, scopeValue := g.GetScope()
+	if len(req.scopes) > 0 { // config-based scopes
+		params.Add(scopeName.String(), strings.Join(req.scopes, " "))
+	} else { // default scopes
+		params.Add(scopeName.String(), scopeValue.String())
+	}
+
 	params.Add("client_id", req.clientID)
 	params.Add("response_type", responseTypeCode.String())
 	params.Add("redirect_uri", req.redirectURI)


asciimoo:
Thanks for the fix. Google does require the scope parameter, but this implementation ignores req.scopes, which contains the configured server.oauth.google.scopes values. Could you merge those configured scopes with the required email and profile defaults? Otherwise additional Google scopes silently have no effect.

jrocketfingers:
Something like this? I updated the PR.
```	scopeName, scopeValue := g.GetScope()
	if len(req.scopes) > 0 { // config-based scopes
		params.Add(scopeName.String(), strings.Join(req.scopes, " "))
	} else { // default scopes
		params.Add(scopeName.String(), scopeValue.String())
	}
```
Feel free to shoot a specific suggestion, or take over the PR.

jrocketfingers:
Oh, and to clarify why I overrode instead of merging -- in case that _anything_ changes, and Google starts requiring something completely different, the user has an **escape hatch** to override with their own config.

asciimoo:
Thank you, I've fixed it in  0b4115e, because there was issues with the GitHub OAuth solution as well.

Thanks again for your contribution.

jrocketfingers:
Yeah, I asked the same thing at the beginning, just didn't wanna make a broad PR for a first contribution. Happy to see you're a good sport, and thanks for Hister, loving it so far 🙌

Issue #534, opened 2026-07-13
[Feature] Disk stats for domains


MitPitt:
With all pages being indexed by default, some websites really explode my storage due to having excess text content.

It would be nice to have a disk stats page, grouped by domain, so I could review and prune unwanted domains, once they pile up. Something like:
```
spammy.com - 567MB (250 pages indexed)
example.com - 100MB (67 pages indexed)
...
```

PerpetualBeta:
# Storage per domain: analysis and design

Notes for #534, written after #706 was closed. Everything with a number in it was measured on a
real index of 2110 documents, not estimated. The probe programs are throwaway; the method is in the
text so any figure can be reproduced.

## 1. What the issue asks for

The report shows domain, size and page count, largest first, so the user can find the sites to
prune. The example in #534 sorts by size. The purpose is pruning, so the order matters more than the
precision of any single row.

## 2. Where the bytes are

Measured on my own index (2110 documents):

| Component | On disk | Attributable to a domain |
|---|---|---|
| `index.db` (Bleve) | 89 MB, of which 78 MB in 4 zap segments | No |
| `data/html` | 89.1 MB in 2083 files | Yes |
| `data/favicon` | 11.2 MB in 433 files | Yes |
| Total | ~190 MB | ~53% |

Two facts follow.

First, the report can account for about half of what hister uses. The other half is the Bleve index,
which holds the stored `text` and `title` fields with term vectors, and which has no per-document
accounting. #534 is about text content that fills the disk, and the text is the part that cannot be
attributed. The report must say so rather than imply that its total is the whole footprint.

Second, blobs are gzip-compressed in the content-addressed store, so a file size is the compressed
size. That is the correct number for a storage report, because it is what a deletion releases, but
the column is "bytes on disk", not "size of the HTML".

Two conditions the report must survive: `disable_previews` stores no HTML at all, which leaves
almost nothing to attribute; and the index can reference blob files that are gone. My index has 3
such references out of 2519.

## 3. Bleve mechanics that decide the design

### 3.1 `SearchAfter` is not a seek

`NewTopNCollectorAfter` keeps the cursor as a document and `Collect` walks every matching document
each request, comparing each one against the cursor (`search/collector/topn.go`). Paging the whole
index therefore costs a multiple of the square of the document count. This is what made #706
quadratic. `Indexer.Iterate` uses the same loop safely because its caller asks for one bounded page.

### 3.2 Facets are one full pass, and return the top N only

[...]

PR #562, opened 2026-08-01
[fix] submit pending page changes on tab close

amosbird:
After a soft navigation, the page title may update after the initial submission. The existing polling corrects this, but only if the tab stays open long enough.

This adds a `visibilitychange` handler that re-extracts the page when the tab is hidden or closed. If the content differs from the last submission, it is sent to the server. No MutationObserver, no dirty tracking, no fixed delays.

Also includes title in the polling change detection so title-only updates are not missed.

Verified with `npm run build -w @hister/ext` and Prettier.

AI disclosure: Codex assisted with implementation and review.

### DIFF
--- a/webui/ext/src/content/content.ts
+++ b/webui/ext/src/content/content.ts
@@ -38,6 +38,25 @@ if (typeof window.navigation !== 'undefined') {
   window.navigation.addEventListener('navigatesuccess', update);
 }
 
+// Submit the latest page state when the tab is being hidden or closed.
+document.addEventListener('visibilitychange', () => {
+  if (!document.hidden || !d || !isContextValid()) return;
+  let current;
+  try {
+    current = extractPageData();
+  } catch (_) {
+    return;
+  }
+  if (current.html != d.html || current.url != d.url || current.title != d.title) {
+    d = current;
+    chrome.runtime.sendMessage({ pageData: d }, (resp) => {
+      if (resp?.status_code === 406) {
+        skippedUrl = d.url;
+      }
+    });
+  }
+});
+
 function scheduleUpdate() {
   if (updateTimer !== null) {
     clearTimeout(updateTimer);
@@ -107,7 +126,7 @@ function update() {
     console.log('failed to extract page data', e);
     return;
   }
-  if (d2.html != d.html || d2.url != d.url) {
+  if (d2.html != d.html || d2.url != d.url || d2.title != d.title) {
     sleepTime = defaultSleepTime;
     d = d2;
     if (d2.url === skippedUrl) {


asciimoo:
Thanks for your contribution.

What use-case does this PR provide a solution for? What is the benefit of capturing previous/intermediate titles?

amosbird:
The goal is not to retain previous/intermediate titles, but to correct them.

On SPA sites such as GitHub, `navigatesuccess` can fire before `document.title` is updated. Hister may then submit the new issue URL with the previous page title (for example, `Build software better, together`). The normal polling may correct it later, but the tab can be closed before the next poll.

This change observes only title mutations, waits 250 ms, and resubmits the complete page with the final title. Submissions are serialized so the earlier extraction cannot overwrite the corrected one.

I encountered this with GitHub issue pages, where history search showed the GitHub homepage/generic title instead of the issue title.

asciimoo:
So basically you would like to update the title as well, when updating a document, right? This does not need a new observer, we just simply have to re-extract the page title when a change is detected (we check if the HTML is changed, so title changes are detected already, without requiring a new observer).

asciimoo:
Ah, I think I understand now the issue. The 250ms wait time is required to give time to the page after a `navigatesuccess` to update the title. The mutation observer isn't even required, because the current change detection is already handling title changes but it has a bigger wait time (10s by default).

I'm asking these questions, because introducing the `MutationObserver` can have significant performance impacts especially on large websites and where title change quickly.
For ordinary SPA navigation this happens once and the performance penalty is insignificant. On sites that frequently change the title, such as unread counters, timers, or notification badges, it could repeatedly extract and upload the entire page. The submission queue could also accumulate large page snapshots if indexing is slower than the title changes.

amosbird:
The HTML comparison does update the title on the next polling check, but there is a window before that check runs.

I reproduced it with the unmodified extension:

1. Opened the GitHub issues list.
2. Navigated to issue #534 by clicking its link (soft navigation).
3. Closed the tab about 3 seconds later, before the 10-second polling check.

Hister stored the destination URL with the previous page's title:

```
URL:   https://github.com/asciimoo/hister/issues/534
Title: Issues · asciimoo/hister
```

This is the same kind of result I originally saw, where an issue URL was stored with `Build software better, together` from the previous GitHub page. When the tab remains open past the polling interval, the existing HTML comparison corrects the title.

The observer is meant to close only this timing window by triggering the existing update path when `<title>` changes. It is limited to title-related mutations under `document.head`; it does not observe arbitrary body changes or replace polling.

If a title observer is not desirable, I am open to a better way to receive this signal promptly. I do not currently see an equally reliable option that avoids guessing a fixed delay; `tabs.onUpdated` would require additional background/content-script coordination.

amosbird:
> I'm asking these questions, because introducing the MutationObserver can have significant performance impacts especially on large websites and where title change quickly.
For ordinary SPA navigation this happens once and the performance penalty is insignificant. On sites that frequently change the title, such as unread counters, timers, or notification badges, it could repeatedly extract and upload the entire page. The submission queue could also accumulate large page snapshots if indexing is slower than the title changes.

That makes sense. I agree that a persistent title observer is broader than needed and could cause unnecessary reindexing on pages with counters or frequently changing titles.

I can simplify this to one extra check shortly after `navigatesuccess`, while keeping the existing polling as a fallback. That should cover the quick-close case without adding an observer or submission queue.

Would that approach work better for you?

asciimoo:
What about simply marking documents as changed/dirty from the mutation observer and sending the changes back to the server only when a `pagehide` or `visibilitychange` event occurs? This solution would catch tab closes but wouldn't overload the server with constant submissions while using no guessed delays.
If this is a working approach, we could use it to detect page changes in general, not just title changes.

amosbird:
That's a much better approach.

Since the extension submits via `chrome.runtime.sendMessage` to the background script, there shouldn't be `sendBeacon` size limits to worry about.

I'll rework the PR to: mark dirty on change detection, submit on `visibilitychange` (hidden) / `pagehide`, and keep the existing initial submission and polling as fallback.

asciimoo:
Awesome! Thanks a lot both for the contribution and for the brainstroming. Let me know if it is available for review.

amosbird:
I've tested it and the fix works as expected. It's ready for review. Thanks for the direction on `visibilitychange` / `pagehide` — much simpler than the original approach.

asciimoo:
How does the current code capture the changes since the last submission? Isn't it require a mutation observer to detect changes and update `d`?
If I understand correctly, only the previously sent document is used again in the new callback. What do I miss?

Also, isn't `pagehide` registered on the wrong target? According to https://developer.chrome.com/docs/web-platform/page-lifecycle-api it should use `window.addEventListener("pagehide", ...)`

amosbird:
My thinking was to skip the observer and just flush on close. But I missed that `d` may still be stale at that point if polling hasn't run yet — I'll add an `extractPageData()` call inside the flush handler to always capture the current state.

I'll also drop `pagehide` since `visibilitychange` already covers tab close.

asciimoo:
> I'll add an extractPageData() call inside the flush handler to always capture the current state

Nice, thanks. I think this is a good approach. It also makes marking the content dirty unnecessary.

amosbird:
I believe it's ready for review now.

asciimoo (approved):
Perfect, thank you!

PR #571, opened 2026-08-06
Bugfix - Add missing label flag

FilipJinstrand:
* Register label flag to cmd flags.
* Add test to make sure all intended flags are available.


asciimoo (approved):
Thanks, LGTM.

asciimoo:
Could you explain when does the label flag missing? The label flag is already registered as a persistent flag on `importCmd` in `cmd/root.go:291`. Every import subcommand inherits it, including file, browser, and all service imports. Cobra merges inherited flags before parsing, so the flag works both before and after the subcommand.
What do I miss?

FilipJinstrand:
> Could you explain when does the label flag missing? The label flag is already registered as a persistent flag on `importCmd` in `cmd/root.go:291`. Every import subcommand inherits it, including file, browser, and all service imports. Cobra merges inherited flags before parsing, so the flag works both before and after the subcommand.
> 
> What do I miss?

My bad seems installing it through winget on windows only have v0.16 as latest which I'm guessing is missing the flag. Building it from source works as intended!

PR #573, opened 2026-08-07
[fix] fully qualify column names in ON CONFLICT statements

### LINKED ISSUE #572: Postgres plus embedding is broken
I'm running Hister on Kubernetes, and today I tried enabling the Postgres integration so I can try out embedding.  
I hit this issue where during indexing, a bunch of errors started showing up in the logs of both Hister and Postgres about an SQL error:

```
2026-08-06T23:42:17Z | ERROR  | server/server.go:1107 > failed to create index error="enqueue embedding: ERROR: column reference \"status\" is ambiguous (SQLSTATE 42702)" URL=<redacted, various URLs>
```

It seems to be because Postgres has stricter requirements when it comes to column disambiguation, which means the fix should (tm) be simple. I'll have a go at a PR shortly.

- Hister version: `ghcr.io/asciimoo/hister:v0.17.0`
- Postgres version: `ghcr.io/cloudnative-pg/postgresql:17.10-standard-bookworm` (via CNPG)

DecentM:
SQLite accepts it, but Postgres requires ambiguous columns to be qualified, otherwise it rejects the request with SQLSTATE 42702 (see #572). I did a search across the server directory and didn't find any other occurrence of this, but I'm happy to revise if other areas need this fix as well. Given the ambiguity I could have gotten the table name wrong as well.  

Also I couldn't find any Postgres tests in here, so no tests :/ (could also expand this PR with test infra, or follow up)

Closes #572 (currently when a user has postgres enabled, embeddings don't work)

AI Usage: I used it to slurp the error messsage from my logs, nothing else.

### DIFF
--- a/server/model/embedding.go
+++ b/server/model/embedding.go
@@ -59,23 +59,23 @@ func EnqueueEmbeddingJob(docID string) error {
 		Columns: []clause.Column{{Name: "doc_id"}},
 		DoUpdates: clause.Assignments(map[string]any{
 			"status": gorm.Expr(
-				"CASE WHEN status = ? THEN status ELSE ? END",
+				"CASE WHEN embedding_jobs.status = ? THEN embedding_jobs.status ELSE ? END",
 				EmbeddingJobInProgress, EmbeddingJobPending,
 			),
 			"dirty": gorm.Expr(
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Thank you.

PR #574, opened 2026-08-07
Catch if path is empty we dont append a / to the requested route.

FilipJinstrand:
I noticed the api endpoints page was empty, issue caused by appending a / to empty path string. This caused the muxServer to respond with the default / respons meaning the whole default route for SPA making the parsing fail.


asciimoo (approved):
You are right, thanks for the fix.

PR #575, opened 2026-08-08
Update Api-docs layout with small improvements

FilipJinstrand:
Definitely not a frontend guy, but I did some small tweaks to the api docs page to be more on-par with the rest of the UI.

* Contents now scroll on its own
* Scroll to section on click
* Improved mobile nav with auto scroll.

Did this with the help of AI : Kimi code

### DIFF
--- a/webui/app/src/routes/api-docs/+page.svelte
+++ b/webui/app/src/routes/api-docs/+page.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import { onMount } from 'svelte';
+  import { page } from '$app/state';
   import { apiFetch } from '$lib/api';
   import { Badge } from '@hister/components/ui/badge';
   import * as Card from '@hister/components/ui/card';
@@ -36,6 +37,9 @@
 
   let endpoints: APIEndpoint[] = $state([]);
   let loading = $state(true);
+  let activeId = $state(page.url.hash.slice(1));
+  let isScrolling = $state(false);
+  let scrollContainer: HTMLDivElement | undefined = $state();
 
   onMount(async () => {
     try {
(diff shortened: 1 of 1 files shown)


asciimoo:
I like the overall direction, thanks for the contribution. Why did you decided to move the ToC to the left of the screen? There is a  big gap between the ToC and the content on wide monitors.

A few minor notes:

 - The links call `preventDefault()` but never update the URL hash. This removes bookmarkable sections and browser history navigation. Initial hash handling also only highlights an item; it does not scroll after the endpoints load asynchronously.
 - The mobile select is not bound to `activeId`. After manual scrolling it keeps its previous value, and selecting that same endpoint again will not emit a change event. Binding its value to the active endpoint would fix both problems.
 - `isScrolling` is cleared only by `scrollend`. When clicking an endpoint already in position, no scrolling may occur and no event may fire. The observer then remains disabled until another successful navigation. A timeout fallback or simpler observer logic would be safer.

Also, thanks for being transparent about AI usage.

FilipJinstrand:
> I like the overall direction, thanks for the contribution. Why did you decided to move the ToC to the left of the screen? There is a big gap between the ToC and the content on wide monitors.
> 
> A few minor notes:
> 
> * The links call `preventDefault()` but never update the URL hash. This removes bookmarkable sections and browser history navigation. Initial hash handling also only highlights an item; it does not scroll after the endpoints load asynchronously.
> * The mobile select is not bound to `activeId`. After manual scrolling it keeps its previous value, and selecting that same endpoint again will not emit a change event. Binding its value to the active endpoint would fix both problems.
> * `isScrolling` is cleared only by `scrollend`. When clicking an endpoint already in position, no scrolling may occur and no event may fire. The observer then remains disabled until another successful navigation. A timeout fallback or simpler observer logic would be safer.
> 
> Also, thanks for being transparent about AI usage.

Fixed the problems. 
And improved the layout a bit more by making it floating and more centred. 

I choose to not update the mobile nav dropdown to reflect to the current viewed section because I feel it looks quite goofy when scrolling past sections and it updates with it. Maybe could add a delay to it and it would look better.

asciimoo:
Great, works well, looks nice.

One last thing: Please use the Svelte state handling instead of `hstory.pushState`/`history.ReplaceState`: https://svelte.dev/docs/kit/$app-navigation#pushState

asciimoo (approved):
Great, thanks!

PR #583, opened 2026-08-15
[enh] add a config option for the profiling endpoints

danmackinlay:
The `/debug/pprof` endpoints are only registered when `log_level` is `debug`, which is too noisy to leave on for a long running indexer. Profiling a memory issue means choosing between the profile and readable logs.

Adds `app.pprof`, defaulting to false. The endpoints keep the existing token/admin authentication wrapper, so this does not change what is exposed without authentication.

Prepared by Claude Code, reviewed by a human.


asciimoo:
Thanks for the contribution.
If we have an explicit config option for the profiler, I'd use only that to activate it. No need to turn it on with the debug log level.

Wouldn't be better to call it `profiler` instead of `pprof`? Also, could you update the configuration documentation with these changes?

danmackinlay:
OK, renamed to profiler, dropped the debug-log-level path so that it's the only switch, and added it to `configuration.md`.

asciimoo:
Great, thanks. Please fix the failing test and it is ready to merge.

asciimoo (approved):
LGTM. Thanks.

PR #584, opened 2026-08-15
[enh] restrict language detection to configured languages

danmackinlay:
`lingua` caches decompressed n-gram models in package level maps that are never evicted. Consequence: every language it is asked about, at every n-gram order it is asked for, stays on the heap for the lifetime of the process. The detector is only used when indexing, so this does not show up in an idle heap profile. Under indexing load it was 362MB of a 392MB heap on my instance.

This PR adds two settings, both defaulting to current behaviour:

```
    indexer.languages                    ISO 639-1 codes, empty means all
    indexer.language_detection_accuracy  "high" (default) or "low"
```

Loading langauges is lazy, so naming the languages keeps the excluded models from ever being decompressed.

Low accuracy restricts lingua to its trigram models. lingua already selects trigrams alone once a document has 120 or more letters, which covers essentially every real page, so detection on full pages is unaffected. Below that it also consults the unigram, bigram, quadrigram and fivegram models, and those then stay loaded for the life of the process. So the largest model tiers were being pinned by pages with very little extracted text, and then never used again. 

Implementation detail: the `initializeIndexer` being a list was getting unwieldy, so the settings are passed as an `indexOptions` struct instead. I can set it back to a list if desired?

Measured on a 2.6GB index, timing backfill plus file indexing to completion, with semantic search enabled and embedding f. The middle row isolates the two settings from each other; the lingua column is that package's share of
`inuse_space`, which does not depend on how much of the workload had finished:

                                   lingua heap   settled footprint
    30 languages, high accuracy          377MB               739MB
    [en de nl id], high accuracy          77MB               164MB
    [en de nl id], low accuracy         3.58MB               105MB

It seems naming the languages is the big win but the accuracy setting helps too — The 77MB to 3.58MB step down is using the same four languages either way, so that difference is the n-gram tiers that only short documents use. 

Naming languages narrows what detection can return, so Latin script text in an unlisted language lands in the default index instead of its own. Those documents stay searchable. Someone who wanted the wide detector could set only
`language_detection_accuracy: low` and leave `languages` empty.

No reindex is needed and nothing is orphaned by this: `initializeIndexer` still opens every `index_*.db` it finds and adds it to the alias, so an existing `index_fr.db` stays searchable after `fr` drops off the list. The setting only narrows what language newly indexed documents can be detected as.

Written with Claude Code, which produced the diagnosis, the measurements and the patch under my direction and multiple revisions. I understand the updates, I believe, but am a Go noob, so could be mistaken.

### DIFF
--- a/cmd/service_import.go
+++ b/cmd/service_import.go
@@ -258,7 +258,10 @@ func newServiceImportRuntime(cmd *cobra.Command) (*serviceImportRuntime, error) 
 	clientOptions := append([]client.Option{client.WithTimeout(0)}, targetUserIDClientOptions(cmd, global)...)
 	languageDetector := document.LanguageDetector(document.NewNullLanguageDetector())
 	if cfg.Indexer.DetectLanguages {
-		languageDetector = document.NewLanguageDetector()
+		languageDetector = document.NewLanguageDetectorFor(
+			cfg.Indexer.Languages,
+			cfg.Indexer.LowAccuracyLanguageDetection(),
+		)
 	}
 	cfg.Crawler.UserAgent = UserAgent
 	applyCrawlerBackendFlags(cmd)
--- a/config/config.go
+++ b/config/config.go
@@ -545,9 +545,10 @@ func CreateDefaultConfig() *Config {
 			MaxBatchBodySize: DefaultMaxBatchBodySize,
 		},
 		Indexer: Indexer{
-			DetectLanguages: true,
-			KeepStopwords:   false,
-			MaxFileSize:     1,
+			DetectLanguages:           true,
+			KeepStopwords:             false,
+			MaxFileSize:               1,
+			LanguageDetectionAccuracy: LanguageDetectionAccuracyHigh,
 		},
 		Crawler: CrawlerConfig{
 			Backend: "http",
--- a/config/config.go
+++ b/config/config.go
@@ -119,6 +119,53 @@ type Indexer struct {
 	KeepStopwords   bool         `yaml:"keep_stopwords" mapstructure:"keep_stopwords"`
 	Directories     []*Directory `yaml:"directories" mapstructure:"directories"`
 	MaxFileSize     int64        `yaml:"max_file_size_mb" mapstructure:"max_file_size_mb"`
+	// Languages restricts detection to the given ISO 639-1 codes. An empty list
+	// detects every supported language. Each language kept here holds its
+	// detection models on the heap for the lifetime of the process.
+	Languages []string `yaml:"languages" mapstructure:"languages"`
+	// LanguageDetectionAccuracy is "high" or "low". Low restricts lingua to its
+	// trigram models, which is what it already uses on its own for text longer
+	// than 120 characters.
+	LanguageDetectionAccuracy string `yaml:"language_detection_accuracy" mapstructure:"language_detection_accuracy"`
+}
+
+const (
+	LanguageDetectionAccuracyHigh = "high"
+	LanguageDetectionAccuracyLow  = "low"
+)
+
+// LowAccuracyLanguageDetection reports whether lingua should run in its reduced
+// memory mode.
+func (i Indexer) LowAccuracyLanguageDetection() bool {
+	return i.LanguageDetectionAccuracy == LanguageDetectionAccuracyLow
+}
+
+// Validate checks what can be checked without knowing which languages the
+// detector supports. The codes themselves are validated in server/document,
+// which owns that list; config keeps no dependency on it.
+func (i Indexer) Validate() error {
+	switch i.LanguageDetectionAccuracy {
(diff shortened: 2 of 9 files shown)


danmackinlay:
Hold on I borked the short string handling. don't bother reviewing, I'll prepare a new one

danmackinlay:
OK, that's less awful. Good to review now.

asciimoo:
Wow, great changes. Both the new language features and the introduction of `indexOptions` are super useful. Thanks for the detailed description & measurements & being transparent about AI usage.

Few questions/notes:

 - The new settings are ignored by service imports. The import runtime (https://github.com/asciimoo/hister/blob/00cfc95b199015a945af009a65158f9d2b6e3d1b/cmd/service_import.go#L259-L262) still calls NewLanguageDetector().
 - I think, the documented behavior is inaccurate. The description says an unlisted language reaches the default index, but doesn't lingua classify it as one of the configured languages?
 - `CreateDefaultConfig`, the website configuration table, examples, and language detection documentation do not include the new config fields.

danmackinlay:
Thanks — and you are right about the classification, my description was wrong.

I checked it. With `languages: [en, de, nl, id]`:

    French    -> en
    Spanish   -> en
    Russian   -> unknown
    Japanese  -> unknown

So Latin script text in an unlisted language is classified as the "nearest" configured language, not a default one. Text only reaches the default index when its script rules out every configured language, as Cyrillic and Japanese do against a Latin-only list.
French getting English stemming is worse than what I claimed in the description, and far from great.

So: data-driven design. Same corpus (88927 chunks embedded), lingua's share of `inuse_space` alongside `Physical footprint`:

                                    lingua heap    footprint
    30 languages, high accuracy          377MB        739MB
    30 languages, low accuracy          18.5MB        107MB
    4 languages, high accuracy            77MB        164MB
    4 languages, low accuracy            3.6MB        105MB

Oh of course, these are _multiplicative_, not additive. 🤦
I didn't do the right experiments in the original PR.
`language_detection_accuracy: low` is doing more than I claimed.
Narrowing the language list on top of it buys about 15MB more, and the footprints are the same within noise.
So as it stands the allowlist permits the misclassification but doesn't gain us much.

My corpus is almost entirely Latin script, so I checked whether that was distorting the result.
Heap size for a detector over all supported languages, fed progressively more diverse scripts:

    detector built, nothing detected      1.0MB
    + latin text                         18.0MB
    + russian                            20.0MB
    + japanese                           20.0MB
    + greek                              20.0MB

The shared (i.e. latin) script is the hog: 20 of the 30 are Latin, so they all need n-grams to tell apart.
Models load lazily, and languages with a script no other configured language uses never load one: `detectLanguageWithRules` identifies them and returns before the n-gram lookup. Russian costs 2MB because Cyrillic is shared with Bulgarian. So a multilingual corpus makes the allowlist worth less rather than more; most of the other languages are “for free”.

Given that, would you prefer `languages` to control index routing only, and leave detection running over all supported languages? A French page would then be detected as `fr`, not be in the list, and go to the default index with the neutral analyzer, instead of being stemmed as English.
It still avoids creating a spread of nearly empty per-language indexes, which was the other thing I was after and it seems memory-affordable.

Alternatively, should we drop `languages` entirely and keep only the accuracy setting?
That is less intrusive, but possibly nearly as good.

Meanwhile I have fixed the other two:

- `service_import.go` now builds its detector from the config instead of calling `NewLanguageDetector()`.
- `CreateDefaultConfig`, the example configuration and the configuration table now include both keys, and the reindex note in the Language Detection section is corrected — neither new setting needs one. The rest of that section we can do after answering above question.

asciimoo:
Thanks for the thorough summary.

The low accuracy memory footprint reduction is impressive. I'm thinking about making it the default instead of the high accuracy. (Of course, it is a bit off topic and the accuracy differences should be measured before deciding)

> Alternatively, should we drop languages entirely and keep only the accuracy setting?

If i understand correctly, the benefit of having the language list configurable is the reduction of index database folders and ~15MB less RAM usage for 30 vs 4 languages. In the mean time all the results with other languages are going to be analyzed/parsed with the general indexer, so search becomes less accurate (same as language detection turned off).
Another aspect is the introduced code complexity.

I don't see this as a significant performance optimization, it's more like a resource usage min-maxing. I'd only keep this config option if the pros and cons are well documented and users are aware of the limited impact of the feature.

What do you think?

> Meanwhile I have fixed the other two

<3

asciimoo:
@danmackinlay What about using `lingua-go`'s `ComputeLanguageConfidenceValues` to identify languages instead of the `DetectLanguageOf` function. Using `ComputeLanguageConfidenceValues` we could decide to classify a language as `unknown` instead of forcing the automatic language resolution to classify the text to one of the configured languages. This way it would make much more sense your original idea - making the supported languages configurable.

What do you think?

PR #585, opened 2026-08-15
Speed up Docker builds and move publishing to Bake

4evy:
Docker builds were repeating a lot of work across the release, root, and debug images

This reworks the Dockerfile to cache npm and Go dependencies, share runtime layers, trim
the build context, and pin base images and yt-dlp downloads

The same targets now live in docker-bake.hcl, so local builds and CI use the same setup. Compose can also reuse the published image and registry cache instead of starting cold

On the CI side, trusted PRs get an amd64 build and Compose smoke test. Merges to master
publish multi-platform images with cache reuse, provenance, and SBOMs. Tagged releases
use a separate no-cache workflow so they always get a clean build

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,68 +1,193 @@
 name: Publish Docker Image
 
 on:
+  pull_request:
+    paths:
+      - .dockerignore
+      - .github/workflows/docker-publish.yml
+      - .github/workflows/docker-release.yml
+      - compose.yml
+      - docker-bake.hcl
+      - Dockerfile
+      - go.mod
+      - go.sum
+      - '**/*.go'
+      - package.json
+      - package-lock.json
+      - webui/app/**
+      - webui/components/**
   push:
-    branches: ['master'] # Trigger on pushes to master branch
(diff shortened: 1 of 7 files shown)


.github/workflows/docker-publish.yml
```diff
@@ -1,68 +1,193 @@
...
 jobs:
-  build-and-push:
-    name: build and push (${{ matrix.target }})
+  build:
+    name: Build Docker images
+    if: >-
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.author_association == 'OWNER' ||
+      github.event.pull_request.author_association == 'COLLABORATOR' ||
+      github.actor_id == '49699333' ||
```
4evy, line 40: 49699333: dependabot

.github/workflows/docker-publish.yml
```diff
@@ -1,68 +1,193 @@
...
+      github.actor_id == '41898282'
```
4evy, line 41: 41898282: github actions bot

asciimoo:
Nice, thanks! Let's see how it performs.

PR #586, opened 2026-08-15
[mod] build the TUI on shared interaction foundations

4evy:
The interface had grown parallel ideas about tabs, menus, shortcuts,
and client behavior, so every new interaction had to keep several
layers in sync

Give those concepts one home, teach the client about labels, previews,
and complete rules, and let each workspace present the keys that
matter there


asciimoo (approved):
Nice, thanks.

PR #593, opened 2026-08-17
[fix] decode opened history responses

4evy:
The history endpoint returns an envelope rather than a bare item list


asciimoo (approved):
Thanks. It would be even better in the future to support both `opened` and the default `visited` history items, just like on the web interface.

PR #594, opened 2026-08-17
[fix] preserve rules when prioritizing results

4evy:
Context-menu prioritization can run before the Rules tab has loaded

Fetch the latest server rules before appending so skip and versioning rules are not overwritten by an empty local cache


asciimoo (approved):
Great, thanks.

PR #599, opened 2026-08-18
[mod] model responsive TUI workspaces and navigation

4evy:
Centralize workspace geometry, navigation targets, and rules metadata in the model. Rendering and input handlers now share state instead of relying on duplicated fixed-row calculations

### DIFF
--- a/client/rules.go
+++ b/client/rules.go
@@ -24,8 +24,12 @@ func (c *Client) FetchRules() (_ *RulesResponse, err error) {
 	return &data, err
 }
 
-func (c *Client) SaveRules(skip, priority, versioning string) (err error) {
-	formData := url.Values{"skip": {skip}, "priority": {priority}, "versioning": {versioning}}
+func (c *Client) SaveRules(skip, priority string, versioning ...string) (err error) {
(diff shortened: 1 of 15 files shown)


asciimoo:
Thanks.

Few notes:
 - Text moved from AddInputs into AddText, while the success handler still clears only AddInputs
 - Deleting a search result no longer refreshes the results or reports failure
 - Mouse tab switching is completely disabled:` tabBar` only examines `m.TabTargets`, but nothing in the pull request tree ever populates that slice
 - The new workspace viewport is never connected to rendering or resize handling. The Add tab wheel modifies `m.Workspace` but `MainView` still renders raw content and truncates it from the top.  Long History and Rules views remain clipped, and Add scrolling has no visible effect.
 - Theme picker mouse selection no longer works
 - `Client.SaveRules` introduces an unnecessary source compatibility break. `SaveRules` changes the optional variadic versioning argument into a required third argument. Existing external callers using two arguments will stop compiling. All internal callers can pass three arguments without removing the compatible signature.

asciimoo (approved):
Great, thanks!

PR #608, opened 2026-08-23
[enh] open search results in a readable details pane

4evy:
Add a focused details view for reading full previews and managing the
selected result. Wide terminals use a split pane, while narrow terminals
give the preview the full workspace

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -92,6 +92,28 @@ func New(baseURL string, opts ...Option) *Client {
 	return c
 }
 
+// FetchConfig retrieves capabilities from the server the client is connected
+// to. This avoids assuming that local configuration describes a remote server.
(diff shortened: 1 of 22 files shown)


asciimoo:
This is fantastic, thank you!

Few notes:

 - Nested overlays corrupt the details return state. Reproduction: open details, open Help, close Help, then press `v` or `esc`. The pane disappears, but the model remains in details state. Preview controls no longer behave correctly.
 - Document text is not wrapped. Long text is clipped instead.
 - Preview loading can flood the server. Rapid dragging or key repeat can create many concurrent requests and spinner loops. It would be good to debounce the requests and do it in series.
 - Semantic settings come from the local client configuration. The TUI should obtain capability and weighting values from `/api/config`.

asciimoo (approved):
Great, thanks.

PR #610, opened 2026-08-23
[mod] migrate terminal interfaces to Charm v2

4evy:
Move command output and the TUI to Bubble Tea, Bubbles, and Lip Gloss
v2. Adapt key, mouse, paste, clipboard, viewport, and declarative view
handling to the new APIs

Query the terminal background for automatic themes, compose overlays at
the cell level, and route styled command output through profile-aware
writers so pipes and `NO_COLOR` stay free of escape sequences


asciimoo (approved):
Thanks!

PR #615, opened 2026-08-23
test: relax yt-dlp fixture timeout

4evy:
The fake yt-dlp process can take more than five seconds to start when the full test suite runs in parallel, causing the extractor to kill it and report a false failure

Use a generous timeout for fake binaries so the tests measure extractor behavior instead of machine load. This does not change the production timeout


asciimoo (approved):
Thanks!

PR #618, opened 2026-08-23
[enh] inherit the terminal color palette by default

4evy:
The TUI currently resolves its default `auto` mode to a bundled Hister theme

That replaces the terminal profile's foreground, background, and ANSI colors even when
the user hasn't chosen a theme

Make a new `terminal` mode the default. It leaves normal text and the screen background
unset while semantic accents use the terminal's ANSI palette

Keep `auto`, `dark`, and `light` as opt-in Hister themes, and expose all four modes
through Settings and the theme picker


asciimoo (approved):
Thanks!

PR #619, opened 2026-08-23
[enh] make TUI focus and feedback explicit

4evy:
The search view could highlight a result while the query input still owned the keyboard.
Enter could then open that result, while mouse-wheel navigation could move the selection
without moving keyboard focus

Empty responses could also leave result focus with nothing to select, including beneath
nested overlays

This aligns selection, keyboard routing, mouse interaction, and footer hints with the
control that actually owns focus

Search responses no longer steal input focus, empty results return focus to the query,
and query suggestions remain actionable. It also adds clearer empty, offline, notice,
and workspace status states and keeps previews and overlays usable on narrow terminals

### DIFF
--- a/cmd/tui/component/keymap.go
+++ b/cmd/tui/component/keymap.go
@@ -14,13 +14,17 @@ import (
 
 	"charm.land/bubbles/v2/key"
 	tea "charm.land/bubbletea/v2"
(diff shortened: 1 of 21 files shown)


asciimoo (approved):
Great, thanks!

PR #108, opened 2026-02-24
[enh] add safari extension

FlameFlag:

PR #632, opened 2026-08-24
Allow safari-web-extension:// origin in CSRF allowlist

rolisz:
I'm using the Safari extension from #108. The CSRF check didn't let it through by default, so this PR adds a check for the Safari specific origin, similar to how Chrome and Safari are handled. 

The extension now works (and I'm aware of all the hoops I'll have to jump through to use it unless it's signed).

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -358,6 +358,10 @@ func withCSRF(handler endpointHandler) endpointHandler {
 				handler(c)
 				return
 			}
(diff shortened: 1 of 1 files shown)


asciimoo (approved):
Thanks!

asciimoo:
This is a good temporary workaround while we don't have an "official" safari extension.

Issue #49, opened 2026-02-15
Safari Support


ellieayla:
I'd like to see Safari supported. The chrome v3 manifest can be converted to a Safari Web Extension ala https://developer.apple.com/documentation/safariservices/packaging-a-web-extension-for-safari which _just works_, albeit unpolished. The extension cannot successfully `POST /add` though. Safari does a pre-flight CORS `OPTIONS /add` request, which fails with http/500 today.

I put together a quick draft PR #46 with server changes to add the minimum required CORS headers to allow the Safari web extension to successfully add to history. This is likely not the best way to add CORS headers, nor is the Safari web extension built or packaged properly. This PR is just a POC to demonstrate a possibility.

ellieayla:
(I'm willing do manual tests of the form "hey, does the server at commit xxxx work with safari?" in the meantime.)

FlameFlag:
I should've clarified this in that post, but you have to do these steps _**on each restart of Safari**_... Unless I'm missing something? 🤔

> Can you think about this a seconde time ?

I'm willing to re-think this if you (or anyone else) could also help out too

> Apple Developper account, the friction disappears

Sadly those are pretty pricy to justify, just for an extension

ticky:
I have an Apple Developer account and I’d be willing to help out the project by publishing a Safari extension (with a zero data collection policy and non-upstream modifications chiefly to the build pipeline to support packaging).

Do you have any requests for how I ought to package/name it? Should I make it clear it's quasi-unofficial?

asciimoo:
@ticky thanks for your offer, it would be great.

> Do you have any requests for how I ought to package/name it?

No. Hister safari extension?

> Should I make it clear it's quasi-unofficial?

Yes, I think it would be good to mention it - just to be fully transparent.

IMNotMax:
This is great news ! Merci ;)

goddamnyouryan:
Any updates?

nburns:
<img width="1303" height="1060" alt="Image" src="https://github.com/user-attachments/assets/6720a784-5d9b-4e94-b166-f5ac3817d835" />

ported to safari and been using it for a while today https://github.com/nburns/hister-safari

feedback/PRs welcome

asciimoo:
@nburns super impressive! What do you think, would it be possible to create the extension artifact on a non mac system?

irdc:
[...]

PR #635, opened 2026-08-24
[enh] allow safari-web-extension origin - refs #49

nburns:
Accept requests from \`safari-web-extension://\` origins on the same subset of endpoints already trusted for \`moz-extension://\` and the packaged Chrome extension. This is the one server-side change needed to make an unofficial signed Safari build of the extension work against a stock Hister server.

Motivation: [#49](https://github.com/asciimoo/hister/issues/49) — you asked contributors to publish/maintain a Safari extension out-of-tree, and offered to accept upstream changes needed on the Hister side. This is that change.

## What changed

- \`server/server.go\` — one additional \`strings.HasPrefix\` on \`Origin\` inside the existing per-path loop in \`withCSRF\`, mirroring the Firefox case. Behavior is unchanged for all existing browsers and origins.
- \`server/origin_test.go\` — new table-driven test covering the three trusted extension prefixes plus a rejected control origin, exercised against \`POST /api/rules\`.

## Verification

- \`go test ./server/...\` passes (full suite).
- Tested end-to-end locally with a Safari build of the extension: enabling the origin allowlist entry lets \`/api/config\`, \`/api/rules\` and \`/api/add\` succeed; without it every request returns 403 as before.

## Scope

Deliberately minimal — no CORS overhaul, no changes to the Chrome/Firefox origin handling, no other files touched. Happy to fold in a stricter allowlist (e.g. gating on a specific Safari extension bundle UUID) if you prefer, but Safari's per-installation UUIDs make a fixed identity check impractical.

The out-of-tree Safari packaging work lives separately and will be published as unofficial per the discussion in #49.

### DIFF
--- a/server/origin_test.go
+++ b/server/origin_test.go
@@ -1,0 +1,47 @@
+// SPDX-License-Identifier: AGPL-3.0-or-later
+
+package server
+
+import (
+	"net/http"
+	"strings"
+	"testing"
+
+	"github.com/asciimoo/hister/server/testutil"
+)
+
+func TestBrowserExtensionOriginsBypassCSRF(t *testing.T) {
+	// A POST to a rules endpoint requires either a valid CSRF token or an
+	// origin that withCSRF explicitly trusts. Requests from the packaged
+	// browser extensions should be trusted; a random cross-origin request
+	// should not.
+	_, handler := newTokenTestServer(t, false)
+
+	cases := []struct {
+		name       string
+		origin     string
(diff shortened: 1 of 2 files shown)


nburns:
created accidentally sorry for the noise, I am working on safari suport though

asciimoo:
>  created accidentally sorry for the noise, I am working on safari suport though

Awesome, thanks!

PR #639, opened 2026-08-25
Add config option to add domains to ytdlp extractor

nsaritzky:
`yt-dlp` has support for [many sites](https://github.com/yt-dlp/yt-dlp/blob/master/supportedsites.md), but the list has caveats (e.g. some sites need an authenticated session to function) and updates frequently, making it unreasonable for hister's ytdlp extractor to match all of them. This PR adds an `extra_domains` config option so that the user can add more sites to the matcher on their own.

Claude Code was consulted for review but didn't get to write any of the code directly.


asciimoo (approved):
Thank you for your contribution. LGTM.

PR #641, opened 2026-08-25
[fix] handle IME composition in search input

koru1130:
For users typing Chinese, Japanese, and Korean, they use Input Method Editor (IME) to compose multiple keystrokes into a character before submitted. During this process, Enter key are used by IME, and it is expected that it won't trigger any action except IME composition itself.

See: https://dev.to/yukimi-inu/why-16-billion-east-asians-are-quietly-raging-at-your-enter-key-handler-1po0

Currently, the home page and search page of this project does not handle IME properly. This PR fixes the relevant keyboard handling and prevents switching pages while composition is still in progress.

### DIFF
--- a/webui/app/src/lib/search.ts
+++ b/webui/app/src/lib/search.ts
@@ -385,6 +385,7 @@ export class KeyHandler {
   }
 
   handle(e: KeyboardEvent, isInputFocus: boolean): boolean {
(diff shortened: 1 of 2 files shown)


asciimoo (approved):
I have limited knowledge about this topic, but the code looks good. Thanks for your contribution.

PR #642, opened 2026-08-25
Add a HackerNews extractor

dlobraico:
Hacker News item pages currently fall through to the generic readability extractor, which keeps the submission text but flattens the discussion.

Unlike the Lobsters markup, Hacker News does not nest its comments: every comment is a sibling row in a single flat table and its depth is carried by the indent attribute on the leading td.ind cell. The extractor rebuilds the tree from that sequence, so both the indexed text and the preview preserve the parent-child hierarchy rather than presenting every comment at the same level.

Match parses the URL instead of prefix-matching it, so pagination and tracking parameters on long threads do not stop a page from being recognised, while lookalike hosts and listing pages are rejected.

Verified against an archived copy of item 8863, which yields the expected title, score, author and a 28k character comment tree.

### DIFF
--- a/server/extractor/extractors/hackernews/hackernews.go
+++ b/server/extractor/extractors/hackernews/hackernews.go
@@ -1,0 +1,298 @@
+// Package hackernews provides an extractor for Hacker News item pages.
+package hackernews
+
+import (
+	"fmt"
+	stdhtml "html"
+	"net/url"
+	"strconv"
+	"strings"
+
+	"github.com/PuerkitoBio/goquery"
+
+	"github.com/asciimoo/hister/server/extractor/sdk"
+	"github.com/asciimoo/hister/server/extractor/textutil"
+	"github.com/asciimoo/hister/server/extractor/urlutil"
+	"github.com/asciimoo/hister/server/sanitizer"
+)
+
+// HackerNewsExtractor turns a news.ycombinator.com item page into searchable
+// text and a readable preview.
+//
+// Unlike the Lobsters markup, Hacker News does not nest its comments: every
+// comment is a sibling row in one flat table and its depth is carried by the
+// indent attribute on the leading td.ind cell. Reconstructing the tree
+// therefore means tracking that number across the row sequence rather than
+// recursing, which is what commentRows below does.
+type HackerNewsExtractor struct {
+	sdk.ConfigSupport
+}
+
+func (e *HackerNewsExtractor) Name() string {
+	return "HackerNews"
+}
+
+func (e *HackerNewsExtractor) Description() string {
+	return "Extracts the submission metadata, self text and full comment tree from Hacker News item pages."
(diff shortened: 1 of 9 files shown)


server/extractor/extractors/hackernews/hackernews.go
```diff
@@ -0,0 +1,288 @@
...
+	doc.Find("tr.athing.comtr").Each(func(_ int, row *goquery.Selection) {
+		depth, err := strconv.Atoi(row.Find("td.ind").First().AttrOr("indent", ""))
+		if err != nil || depth < 0 {
+			depth = 0
+		}
+		text := row.Find(".commtext").First()
+		// Collapsed and flagged comments keep their row but carry no body.
+		// They are still worth a line so the thread shape survives, but an
+		// entirely empty one adds nothing to the index.
+		body := strings.TrimSpace(text.Text())
```
asciimoo: Paragraph boundaries are lost this way. It would be good to keep the formatting of the original comment.

server/extractor/extractors/hackernews/hackernews.go
```diff
@@ -0,0 +1,288 @@
...
+		return sdk.PreviewFallback(err)
+	}
+
+	s := parseStory(doc)
+
+	var b strings.Builder
+	if s.title != "" || s.link != "" {
+		b.WriteString("<h2>")
+		if s.link != "" {
+			fmt.Fprintf(&b, `<a href="%s">%s</a>`, stdhtml.EscapeString(s.link), stdhtml.EscapeString(s.title))
```
asciimoo, line 218: Relative links disappear from previews. Ask HN, polls, and internal HN links commonly use values such as `item?id=123`. Please resolve the links against d.URL before sanitization.

asciimoo (changes requested):
Great PR. Thanks for your contribution. I've added a few notes, please check it.

server/extractor/extractors/hackernews/hackernews.go
dlobraico: Good point! I grabbed the structure aware helper in the Discourse/Reddit extractors and refactored it into a shared utility so that I could use it here too. Let me know if you'd prefer I copy/paste it again here.

server/extractor/extractors/hackernews/hackernews.go
dlobraico: Fixed!

dlobraico:
Thanks for the feedback! These should both be resolved now.

asciimoo:
@dlobraico thanks! Please fix the failing github workflows and remove claude from the commits, then it is ready to be merged.

dlobraico:
Sorry about that, should be good now!

asciimoo:
Great, thanks! The test that is still failing is independent from the PR

PR #644, opened 2026-08-25
Add native macOS launchd setup for non-Nix users

sathwick-p:
Added a native macOS launchd setup for users who run Hister directly without Nix or Docker. The helper installs a per-user LaunchAgent that runs hister listen, starts at login, restarts after unexpected exits and shutdowns.
I've tested the setup manually on my mac couple of times and also have updated the webui as well.

AI Disclosure : I've used gpt-5.6-sol for implementation feedback, review, and tests.

### DIFF
--- a/contrib/launchd/hister-launchd.sh
+++ b/contrib/launchd/hister-launchd.sh
@@ -1,0 +1,355 @@
+#!/bin/sh
+
+# Manage a per-user macOS launchd agent for Hister.
+# The script keeps Hister's data and logs in the user's home directory and
+# never requires root privileges. It is safe to run from a source checkout or
+# with an installed binary selected through HISTER_BIN.
+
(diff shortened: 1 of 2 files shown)


nburns:
ITT just a plist with some instructions as a comment would be better, a big script is kinda hard to understand

There're conventions about where files go on install so it's quite possible that those wouldn't even need to be templated i.e. /usr/local: https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard bit different on a mac but the general idea is the same

sathwick-p:
In that case the user would have to copy to the LaunchAgents directory and manually run launchctl bootstrap commands. The current helper also handles custom paths, config/data locations and service management removing these will add more setup and maintenance overhead for the user.

Issue #643, opened 2026-08-25
Respectful client / anti-bot care: rate limiting


nburns:
I noticed yesterday that I was seeing more cloudflare rejections for "bot behavior", wondering if its related and if you would accept a PR with some rate limiting in it?  Imagining it would be nice to limit requests globally and per host with a cooldown/backoff for server errors.

Suspecting its the double fetch of some of the pdfs that triggered this, but would be nice in general to be a respectful client of other servers.

nburns:
Also related to my thinking about importing iCloud tabs and Safari's Reading List: https://github.com/nburns/hister-safari/issues/1#issuecomment-5416622787

asciimoo:
By "client" are you referring to the extension or the built-in crawler?

nburns:
The crawler

asciimoo:
I agree, the crawler is still very rudimentary.  Better rate limiting would be a great addition, and your contribution would be very welcome.

PR #645, opened 2026-08-26
Make crawler well behaved (local and remote resource usage)

nburns:
from: https://github.com/asciimoo/hister/issues/643

Makes the crawler behave a bit better:
- global/per host rate limits
- easier on ram/cpu
- follow robots.txt
- url normalization
- user agent, with page on docs site
- configuration for new options, also respects old config with a warning
- tests and verified locally

I used claude to generate code, but planned the changes and wrote this myself, happy to make any modifications/updates.

crawler docs:
Screenshot 2026-08-25 at 5 20 49 PM

config docs:
Screenshot 2026-08-25 at 5 29 12 PM

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -132,15 +132,74 @@ type CrawlerCookie struct {
 // CrawlerConfig holds global crawler settings and backend-specific options.
 // Timeout and Delay default to 5s and 0s respectively when zero.
 type CrawlerConfig struct {
-	Timeout        int               `yaml:"timeout"         mapstructure:"timeout"`
-	Delay          int               `yaml:"delay"           mapstructure:"delay"`
-	Backend        string            `yaml:"backend"         mapstructure:"backend"`
-	BackendOptions map[string]any    `yaml:"backend_options" mapstructure:"backend_options"`
-	Proxy          string            `yaml:"proxy"           mapstructure:"proxy"`
-	UserAgent      string            `yaml:"user_agent"      mapstructure:"user_agent"`
-	Headers        map[string]string `yaml:"headers"         mapstructure:"headers"`
-	Cookies        []CrawlerCookie   `yaml:"cookies"         mapstructure:"cookies"`
-	NoRobots       bool              `yaml:"no_robots"       mapstructure:"no_robots"`
+	Timeout           int                            `yaml:"timeout"            mapstructure:"timeout"`
+	Delay             int                            `yaml:"delay"              mapstructure:"delay"`
+	Backend           string                         `yaml:"backend"            mapstructure:"backend"`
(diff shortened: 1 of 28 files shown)


asciimoo:
Wow, this is a super impressive and substantial change. It will take some time to review it thoroughly. One minor change I'd request right away: please move the stats menu item on the app UI to the dropdown menu on the right, as I expect this endpoint to be used less frequently.

nburns:
Appears I have some review of my own to do too that I missed

nburns:
Also got bigger than I initially imagined, so we can split it up if thats more palatable

asciimoo:
> Also got bigger than I initially imagined, so we can split it up if thats more palatable

Yes, please separate out at least the stats feature, if it isn't too much trouble.

JonnieCache:
I've been looking at doing a similar implementation, so I'll give some thoughts. Firstly a few of the trickier to review parts could be replaced with standard primitives:

- `clock.go` is in the standard library at [testing/synctest](https://pkg.go.dev/testing/synctest).

- The job of `lazyBucket` could be done by [x/time/rate](golang.org/x/time/rate).

- The job of `getOrFetch()` on `Robots` could be done by [x/sync/singleflight](golang.org/x/sync/singleflight).

Some other points:

- The app has two crawler loops with duplicated logic, `bfsCrawl()` and `persistentBFS()`, one for the browser ingress and one for the history import. Really these should be merged into one with some adapters or something. There we can add a single request worker pool to take full advantage of the new backoff code.

- Without that last part, this PR will actually make the history import slower, as it's still single threaded and now it's doing a bunch of waiting.

- Sqlite is currently set up for single threaded operation. It needs the WAL and suchlike enabled to work safely with a thread pool (which is already how its used incidentally.)

- Nothing is telling the browser fetchers or the extractors what level of concurrency is happening, and some of them will set up thread pools of their own, sending out a lot more requests which arent controlled by the budget.

Also I agree that this PR is too big.

JonnieCache:
Another observation, this design splits the state for each host across a bunch of data structures, `Scheduler.hosts`, `RobotsCache.cache`, `RobotsCache.hostLocks`, `CircuitBreaker.hosts` and `Budget.perHost`. It might simplify things to have one map of `HostState` structs with everything in one place.

nburns:
Good feedback, I can say it's sorta bordering on slop (my PR that is) so if you want to take it over, I'm happy to hand it off.

nburns:
Going to split into a couple more PRs:
- well behaved crawler as seen by external sites including cookies, rate limiting, etc (my original motivation)
- one for the host resource usage (got greedy/sloppy)

nburns:
closing in favor of a couple smaller PRs

PR #648, opened 2026-08-26
import browser bookmarks (Firefox, Chromium, Ladybird)

### LINKED ISSUE #626: command to import browser bookmarks
I have a highly curated set of firefox bookmarks. Being able to import them into hister would really boostrap it's usage for me and increase my confidence in its results. I don't see in the docs any way to import them (there is a lot).

asjdf:
## Summary

`hister import browser` is history-only (`moz_places` / Chromium `History` / …). This adds bookmarks as a sibling subcommand:

```
hister import browser bookmarks --browser firefox
hister import browser bookmarks --db ~/.config/google-chrome/Default/Bookmarks
```

Sources live under `pkg/browser/bookmarks` (Firefox `places.sqlite`, Chromium-family `Bookmarks` JSON, Ladybird `Bookmarks.json`). They enqueue through the same crawl job as history import. Skip rules apply; there is no extra denylist for a browser's shipped defaults.

`--start-date` is omitted on purpose: it filters `last_visit_date` and would drop bookmarks that have never been opened.

Job IDs:

- `browser-history-import-YYYY-MM-DD`
- `browser-bookmark-import-YYYY-MM-DD`

Fixes #626.

## Test plan

- [x] `go test ./cmd ./pkg/browser/bookmarks/...` (Firefox 140 schema fixture, Chromium/Ladybird JSON)
- [ ] `hister import browser bookmarks` against a local profile

Draft for review. Implementation was assisted in Cursor.


asciimoo:
Nice! Thank you for your contribution.

Currently the bookmark jobs cannot actually be resumed. `chooseBrowserImportJobID` only selects jobs beginning with `browser-import-` prefix. Could you extend the job selector to accept the prefix as argument and make the bookmark import resumable?

Another nice to have idea: adding the bookmark folder as a label to the imported documents. I'm not sure if it can be done without major changes in the codebase. What do you think?

asjdf:
Good call — `chooseBrowserImportJobID` / `browserImportJobs` now take the prefix. Bookmark jobs (`bookmark-import-*`) show up in the resume prompt; history stays on `browser-import-*`. Resume also keeps the job's own default label (`bookmarks`) instead of falling back to `"browser"`.

Pushed on this branch in 1c5881f.

On the command shape: I nested bookmarks under `import browser` because the source is the browser, and history/bookmarks are the two things we pull from it.

- `import browser` stays the old positional history import (compat)
- `import browser history` is the named `--browser` / `--db` twin
- `import browser bookmarks` is the same flags for bookmarks

A sibling `import bookmarks` would put history and bookmarks at different levels. Didn't want that.

Folder as document label: `Document.Label` is one string, and the crawl job has one label for the whole job. `BulkInsertCrawlURLs` also has no per-URL field. So it's not a small add-on — we'd need either one job per folder, or a URL→folder map applied when the document is created. Happy to do the latter if you want it, but I'd keep it out of this PR. `--label` already covers a single job-wide label.

asjdf:
One more thought on the job IDs: instead of a parallel `bookmark-import-*` family, would you rather both sit under `browser-*-import-*`?

- `browser-history-import-YYYY-MM-DD`
- `browser-bookmark-import-YYYY-MM-DD`

Same prefix argument on the selector, just one naming scheme. The current `bookmark-import-*` is only on this branch, so easy to switch if you prefer that.

asciimoo:
> One more thought on the job IDs: instead of a parallel bookmark-import-* family, would you rather both sit under browser-*-import-*?

Yes, I agree, this is a more consistent naming convention.

CloneOfAlex:
## A few additional recommendations

Nice work on the multi-browser bookmark support. A few things worth addressing before merge, beyond the resumability/naming changes already made:

**1. Duplication between `bookmarks.go` and `browser.go` import pipelines**
`importURLGroups`/`multipleChoiceURLGroups` (`cmd/bookmarks.go`) re-implement most of what `importDB`/`multipleChoiceImport` (`cmd/browser.go`) already do: resolve job prefix/label, create-or-resume the crawl job, batch-insert URLs, check `crawlJobHasURLsToCrawl`, spin up the crawler. Recommend factoring a shared helper both commands call with a small kind-specific config, so bug fixes don't need to land twice.

**2. `bookmarkSourceHasName` over-matches on longer input**
```go
n == name || strings.HasPrefix(name, n) || strings.HasPrefix(n, name)
```
The last clause means `--browser firefoxx` (a typo) silently matches `"firefox"` instead of erroring as unknown, since `"firefoxx"` starts with `"firefox"`. Suggest a one-directional prefix check (`strings.HasPrefix(n, name)` only), matching the pattern already used in `historyDBs`.

**3. No cross-group URL de-dup before enqueue**
`UniqueHTTPURLs` dedupes within one store, but URLs aren't deduped across groups (e.g. the same bookmark synced to both Chrome and Firefox). Please confirm `BulkInsertCrawlURLs` enforces uniqueness per `(job_id, url)` server-side, or add a dedup pass across `groups` before enqueueing.

**4. Test coverage gap on the branchier new logic**
Current tests cover `resolveBookmarkStores` for `--db`-only and one rejection case. Missing: autodetection via `Detect()` across multiple profiles, `bookmarkSourceHasName` fuzzy-match edge cases (including #2 above), and `multipleChoiceURLGroups`'s exclude-list parsing.

**5. Minor: Ladybird profile discovery is inconsistent**
`Detect()` globs `~/.config/Ladybird/*/Bookmarks.json` for extra profiles but only checks the single literal path under `~/.local/share/Ladybird/`. Same glob treatment for both roots would be more consistent.

**6. Minor: dead defensive check**
`cmd.Flags().Lookup("min-visit") != nil` in `importDB` looks unreachable today — every current caller registers `min-visit`. Fine to keep as future-proofing, but a short comment on why would help the next reader.

asjdf:
Switched the prefixes as discussed:

- `browser-history-import-YYYY-MM-DD`
- `browser-bookmark-import-YYYY-MM-DD`

Pushed in 1d3b9f3.

asjdf:
Thanks for the notes. Addressed in 1d3b9f3.

- Shared helper: `beginBrowserImportJob` / `enqueueBrowserImportURLs` / `finishBrowserImportJob`. History still streams sqlite rows; bookmarks still start from an in-memory URL list. Job select, resume, enqueue, and crawl now go through one path.
- `bookmarkSourceHasName` now matches `historyDBs`: exact or source name starts with the input. `--browser fire` still hits firefox; `--browser firefoxx` is unknown.
- Cross-group dedup: `BulkInsertCrawlURLs` already has `uniqueIndex(job_id, url)` and `OnConflict DoNothing`, so a second Chrome/Firefox copy of the same URL is dropped server-side. Left that as-is.
- Tests: HasName edges (including the firefoxx typo), Detect across multiple profiles, Ladybird dual-root glob, and exclude-list parsing.
- Ladybird now globs both `~/.config/Ladybird/*/Bookmarks.json` and `~/.local/share/Ladybird/*/Bookmarks.json`.
- Left the `min-visit` nil check; added a comment that current callers all register the flag and the lookup is for a future import path without it.

asciimoo:
Great, thanks!
Few minor notes left:
- I think Ladybird places named profiles under `Ladybird/Profiles/<identifier>` but the detector searches `Ladybird/*/Bookmarks.json`
- The documentation still names jobs `bookmark-import-*` and `browser-import-*`

asciimoo (approved):
Awesome, thank you!

asciimoo:
Please fix the lint & formatting & tests and it is ready to be merged.

asjdf:
Addressed in `0044b26`.

The earlier Ladybird globs (`Ladybird/*/Bookmarks.json`) were from older notes, not a live install. I installed Ladybird on Linux and the real layout is `Ladybird/Profiles/<id>/Bookmarks.json` (Flatpak remaps that under `~/.var/app/org.ladybird.Ladybird/...`). Bookmarks are under config, history under data, so they are not siblings.

Detect and `getDBPaths` now follow that, plus the documented macOS `Application Support` paths. I have not been able to check macOS or Windows installs — help verifying those would be useful.

asciimoo:
Seems like the tests/checks are still failing here on Github

asciimoo:
@asjdf could you rebase the PR and fix the tests?

asjdf:
Sorry for missing your message. I’ll look into it again.

asciimoo:
Oh, sorry, 3484623 is on me.

Looks good except one thing: The new history `--db` command loses existing database detection behavior. It guesses from filenames, misidentifying Safari’s `History.db`. Would be better to reuse `detectHistoryTable`.

Please fix this and it's ready to merge.

asjdf:
Addressed in 74fdfb9.

`historyTableFromPath` now calls `detectHistoryTable`, so Safari's `History.db` is no longer treated as Ladybird when `--db` is used without `--browser`.

asciimoo (approved):
Great, thanks!

k4lizen:
thanks!

PR #649, opened 2026-08-26
fix: keep vectors after reindex errors

### LINKED ISSUE #354: Rewrite semantic search reindex to be fail safe
The Bleve reindex and the vector store handle failures asymmetrically.

Bleve: the new indexes are built under basePath/reindex, and the old indexes are only removed once the rebuild loop completes successfully. A mid-loop failure leaves the old indexes untouched, so keyword search keeps working with the pre-reindex data.

Vector store: vs.Clear() wipes every embedding before the loop runs. A mid-loop failure (embedding endpoint down, panic, SIGTERM, a single embed error returning err from b.Add, etc.) leaves the vector DB empty with no way back except another full reindex.

The inline comment ("no temp-dir / rename dance is needed because it is a separate file") justifies the in-place approach, but being a separate file arguably makes temp/rename easier, not unnecessary. Worth mirroring the Bleve approach, or at least calling this out in the docs so operators understand a failed reindex equals an empty vector DB.

KevinAndrewDong:
Fixes #354

A failed semantic re‑index may empty the vector store after existing data gets cleared. All vector updates are now staged. Changes only go live once re‑index completes successfully, and get rolled back on embedding errors or job cancellation.

The existing vector and Bleve indexes stay usable until commit. Embedding errors are propagated upstream. Regression tests cover failure, cancellation, rollback, and successful‑commit scenarios.

Codex assisted with implementation and review. I’ve reviewed and fully understand the final patch.

### DIFF
--- a/server/indexer/indexer.go
+++ b/server/indexer/indexer.go
@@ -51,11 +51,14 @@ type Indexer struct {
 	indexers          map[string]bleve.Index // default and language specific indexers
 	indexesMu         sync.RWMutex           // protects idx, indexers, and indexesClosed
 	indexCreationMu   sync.Mutex             // serializes index creation, close, and adoption
+	lifecycleMu       sync.Mutex             // serializes Reindex and Close
+	reindexMu         sync.RWMutex           // serializes reindex publication with mutations
 	indexesClosed     bool
 	dir               string
 	data              *dataStore
 	langDetector      document.LanguageDetector
 	reindexInProgress atomic.Bool
+	reindexGeneration atomic.Uint64
 	embedder          *vectorstore.Embedder
 	vectorStore       vectorstore.VectorStore
 	embedCtx          context.Context
@@ -63,6 +66,7 @@ type Indexer struct {
 	embeddingQueue    *embeddingQueue
 	embeddingWorkers  int
 	disablePreviews   bool
+	detectLanguages   bool
 	keepStopwords     bool
 	directories       []*config.Directory
 	maxFileSize       int64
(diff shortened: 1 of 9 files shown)


server/indexer/indexer.go
```diff
@@ -756,34 +833,95 @@ func (idx *Indexer) reindex(ctx context.Context, basePath string, rules *config.
...
 	}
 	replacement, err := initializeIndexer(basePath, detectLanguages, keepStopwords)
 	if err != nil {
-		return err
+		return recoverOld(err)
 	}
 	replacement.maxFileSize = idx.maxFileSize
 	replacement.sensitivePattern = idx.sensitivePattern
 	replacement.semanticConfig = idx.semanticConfig
+	if stagedVS != nil {
```
asciimoo, line 916: This still publishes the new Bleve indexes. The code clears `rollbackPublication` and adopts `replacement`, while restoring the old vector store. That leaves keyword and semantic indexes representing different document sets, contradicting the atomic publication goal. This path should restore both stores, probably through `recoverOld`.

server/indexer/indexer.go
```diff
@@ -582,20 +632,31 @@ func (idx *Indexer) reindex(ctx context.Context, basePath string, rules *config.
 	if err := ctx.Err(); err != nil {
 		return err
 	}
+	idx.lifecycleMu.Lock()
```
asciimoo: Concurrent reindex requests no longer return “already running.” `lifecycleMu` is acquired before `CompareAndSwap`, and the first reindex holds it for its entire duration. A second request waits, reaches the check only after the first clears the flag, then starts another complete reindex.

server/indexer/indexer.go
```diff
@@ -582,20 +632,31 @@ func (idx *Indexer) reindex(ctx context.Context, basePath string, rules *config.
...
 	// TODO store new documents in both indexes while running reindex to guarantee not losing any data.
 	if !idx.reindexInProgress.CompareAndSwap(false, true) {
+		idx.lifecycleMu.Unlock()
 		return errors.New("Reindex is already running")
 	}
+	idx.reindexGeneration.Add(1)
```
asciimoo, line 662: There is a race around `embeddingQueue`. Reindex stops and replaces the queue before acquiring `reindexMu`, while normal mutation paths read the queue under `reindexMu`. Concurrent indexing or deletion can therefore race with `stopEmbeddingQueue`.

asciimoo (changes requested):
Great, thanks for the contribution and for disclosing AI usage. Please check my review notes.

PR #687, opened 2026-08-28
nix: use upstream NixOS module

4evy:
Since Hister is now on nixpkgs, we can just use upstream module


asciimoo (approved):
Great, thanks!

PR #692, opened 2026-08-29
[enh] allow safari-web-extension CORS preflight - refs #49

ad3lre:
Safari's extension popup still sends a CORS preflight (`OPTIONS`) for JSON bodies and `X-Access-Token`. Those requests currently fall through to the SPA handler, so token auth from the unofficial Safari extension fails even though CSRF already allows `safari-web-extension://` origins on a subset of paths.

This adds a small middleware that, only for trusted browser-extension origins on the extension API surface:

- answers `OPTIONS` with `204`
- reflects that origin in `Access-Control-Allow-*`
- skips CSRF on `/api/history` as well (the content script posts search-result history there; Chrome/Firefox already share this origin check)

Trusted origins are unchanged: `moz-extension://`, `safari-web-extension://`, and the packaged Chrome extension id. Other origins are not given CORS headers.

## Test plan

- [x] `go test ./server/ -run TestExtension`
- [x] `OPTIONS /api/add` from a `safari-web-extension://` origin returns 204 with `Access-Control-Allow-Origin` set to that origin
- [x] the same `OPTIONS` from `https://evil.example` does not get those headers
- [x] `POST /api/add` and `POST /api/history` from the Safari origin succeed (201 / 200) with an access token

Assistance: Cursor was used while writing the middleware and tests. The behaviour follows issue #49 and the earlier CORS POC in #46.

### DIFF
--- a/server/extension.go
+++ b/server/extension.go
@@ -46,6 +46,19 @@ func browserExtensionRequest(origin, path string) bool {
 	return trustedBrowserExtensionOrigin(origin) && browserExtensionAPIPath(path)
 }
 
+// extensionCORSAllowHeaders returns the response Access-Control-Allow-Headers
+// value. Safari preflights every extension fetch and lists the actual request
+// headers in Access-Control-Request-Headers; echoing that list keeps
+// user-configured custom headers (reverse-proxy auth, etc.) working. When no
+// preflight header is present, fall back to the headers the extension always
+// sends.
+func extensionCORSAllowHeaders(r *http.Request) string {
+	if requested := r.Header.Get("Access-Control-Request-Headers"); requested != "" {
+		return requested
+	}
+	return "Content-Type, X-Access-Token, X-Hister-Public, X-CSRF-Token, Authorization"
+}
+
 // withExtensionCORS answers CORS preflight and attaches CORS headers for
 // requests that come from a trusted browser-extension origin on the extension
 // API surface.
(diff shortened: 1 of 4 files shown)


asciimoo:
Thanks for your contribution.

@nburns could you help verifying the issue/solution?

ad3lre:
I hope to be a long term contributor on here as i have interest in Hister and supporting its compatibility on Safari as an avid Safari user myself

please let me know if anything goes wrong / if I did.

server/extension.go
```diff
@@ -0,0 +1,77 @@
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		origin := r.Header.Get("Origin")
+		if !browserExtensionRequest(origin, r.URL.Path) {
+			next.ServeHTTP(w, r)
+			return
+		}
+		h := w.Header()
+		h.Set("Access-Control-Allow-Origin", origin)
+		h.Set("Access-Control-Allow-Methods", "GET, HEAD, POST, OPTIONS")
+		h.Set("Access-Control-Allow-Headers", "Content-Type, X-Access-Token, X-Hister-Public, X-CSRF-Token, Authorization, Cookie")
```
nburns: ITT Cookie gets removed by the browser/cant be set from JS so we can remove this?

https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_request_header

nburns (commented):
Looks reasonable to me

server/extension.go
ad3lre: Yea you're right, I'll get rid of it

asciimoo:
Thanks for both the review and the changes.

Another issue: the fixed header allowlist breaks custom headers. The extension supports arbitrary headers for e.g. reverse proxy authentication, but the middleware permits only a fixed set in `Access-Control-Allow-Headers`.

PR #706, opened 2026-08-31
Report storage per domain

### LINKED ISSUE #534: [Feature] Disk stats for domains

PerpetualBeta:
Closes #534.

This change adds a storage report grouped by domain. It shows what the index contains, and which
domains you can delete to release space.

The report is available on three interfaces:

- A `/storage` page in the web UI.
- A `hister stats domains` command.
- `GET /api/stats/domains`.

Issue #534 asks for the domain, the size and the page count. This report also divides the size into
its parts. A domain that holds 500 MB of HTML and a domain that holds 500 MB of indexed text are
different problems. The correction for one is not the correction for the other.

### The sizes are what deletion releases

The stored data is content addressed. Two pages with identical HTML share one file. The report
charges a shared file to no domain.

The alternative is to charge every domain that refers to the file. This makes the columns add up to
more than the disk contains. It also shows space that the deletion of one domain does not release.

There is a result of this rule. A domain can look smaller than the user expects. Thus the report
has a `SHARED` column. Without the column, a domain whose pages are identical to the pages of a
different domain shows zero storage. This looks like a defect.

On one real index of a few thousand pages, shared HTML files were rare. Shared favicons were
frequent, because one icon usually serves a full site.

### `TextBytes` is an estimate

`TextBytes` is the length of the indexed text. It is an estimate of the share of the search index
that the domain uses. It is not a measurement of that share. Bleve does not report the storage of
each document. The index also contains term vectors and postings that no single document owns.

The HTML bytes and the favicon bytes are exact. They are file sizes on disk.

I include `TextBytes` because indexed text is the data that increases, and #534 is about text
content that fills the disk. But it is an estimate in a column adjacent to two exact figures. The
web page and the CLI both show a note about this. Tell me if you prefer a stronger note, or if you
prefer to remove the column.

### Performance

The report reads all the documents in the index. The sizes apply to each document, and there is no
aggregate to read instead.

The report does one pass over the stored fields. It then does one `os.Stat` for each different
blob. It does not open the blobs. Thus a report about 89 MB of stored HTML does not read 89 MB.

The report is still slow on a large index. Thus the CLI sends the request with no timeout. It is
better to wait for the report than to stop it when a default time elapses.

### Authentication

The endpoint always needs authentication. It gives the data for one user when user handling is on.
`TotalByUser` uses the same rule.

A storage report shows the sites that a user visits most, in sequence. Thus the endpoint gives no
data to an unauthenticated caller, also on an instance that is public in other respects.

### Tests

`go build ./...`, `go test ./...` and `golangci-lint run ./...` are clean.

There are four new tests:

- `TestDomainStatsGroupsAndCounts` tests the grouping and the page counts.
- `TestDomainStatsChargesBlobsOnlyToTheirSoleOwner` tests the rule above. Two domains share one
  blob. The test makes sure that the report charges neither domain, and reports the shared bytes to
  both.
- `TestDomainStatsHTMLBytesMatchTheFileOnDisk` compares the exact figures with the file. It does not
  compare them with a second calculation.
- `TestDomainStatsTotalIsTheSumOfItsParts` makes sure that the columns add up.

NOTE: `npm run lint` fails on this branch. It fails in the same way on `master`. Both give 4
dead-code issues, 6 clone groups and 131 health findings, and both start at the same file. The lint
does not report the new files. The maintainability index is 92.9 for both. The lint analyses 884
files on this branch and 877 files on `master`.

### Notes for the review

- The report shows sizes to one decimal place from kilobytes and above. Whole units make a row such
  as "4 KB + 4 KB = 2 MB" look incorrect. Users do not trust a storage report whose columns appear
  not to add up.
- The footer does not add the `SHARED` column. An addition counts the same bytes one time for each
  domain that refers to them.
- An initial version showed the relative size as a bar. I removed it. The distribution is very
  unequal. On a real index, a linear bar made almost one half of the rows less than one pixel wide.
  A square-root scale corrected the width, but then the width was no longer proportional. The report
  now shows a percentage.
- The report groups documents that have no domain as `(unknown)`. It does not discard them. Thus the
  totals describe the full index.
- If a blob file is absent, the report continues and does not count it. The index can stay after its
  data file if a cleanup stops before it is complete.
- The change also adds documentation to `data-lifecycle.md` and `terminal-client.md`, and one entry
  to the navigation menu.


asciimoo:
The implementation is inefficient, inaccurate and does not scale well.

This is a complex problem and the provided solution requires significant changes and more planning in my opinion.

A few remarks (it isn't a complete review):

It would be better to use Bleve facets, to aggregate domains and iterate over each domain independently instead of iterating over the whole index (btw, using `SearchAfter` results in an O(n^2) offset calculation - check how the paging is done in other parts of the app).

There is no paging, all the domains are collected/calculated in one request.

As I see  memory usage grows approximately with the number of documents.

The Bleve size changes are speculative. `len(text)` is input text size, it is not Bleve index storage, and Bleve deletion may not immediately shrink the index file. It also ignores indexed titles, URLs, term vectors, and other structures.
I don't have deep knowledge about Bleve internals, but I know it uses references to text chunks, so if pages share texts, then as long as one remains in the index, the deletion of the others does not reduce the index size significantly.
Probably I'd fully exclude the index size calculations at first.

PerpetualBeta:
Thank you for the time you gave this. You are correct on every point, and I agree with the
conclusion. I approached the problem as a report to write and not as a problem to solve.

I read the code before I replied. Three things I can add:

The paging is worse than an offset calculation. Bleve keeps the `SearchAfter` cursor in the
collector and compares it against each document it visits. It is not a seek. So every page walks
the full matching set again. I copied the loop in `Indexer.Iterate` without asking why that shape is
acceptable there: it returns one bounded page to a caller who asks for a few, and mine walks the
whole index inside one request.

The facet you want already exists. `searchschema` defines a `domains` terms facet on the `domain`
field, and `addFacets` already uses it. A search with `Size = 0` gives the domains and the page
counts with no document iteration. I will start from there.

The memory is in two maps that record which domain refers to each blob. They are keyed by blob, so
they grow with the document count. Iteration for each domain removes the need for them.

I will remove all the index size calculations. You are right that a caveat in the description does
not correct a number in a column adjacent to two exact figures.

I am closing this PR. I will put a plan on #534 for your comment before I write more code. One
question for that plan, which I do not yet have an answer to: a terms facet gives the largest
domains by count, so I must find how to page through all the domains, or agree that the report shows
only the top domains.

Sorry for the review time this used.

asciimoo:
> The paging is worse than an offset calculation. Bleve keeps the SearchAfter cursor in the
collector and compares it against each document it visits

Sorry, you are right. It is the efficient way to iterate, I mixed it with defining offsets.

> The facet you want already exists. searchschema defines a domains

Exactly!

> I will remove all the index size calculations.

It would simplify the implementation a lot. We can introduce it later.

> Sorry for the review time this used.

No worries, this PR made me think about this topic in more detail, that's how I came to a conclusion that this is a more complex topic than it seems at first. So thank you for the contribution, it was inspiring, even if it won't be part of the codebase in its current form.

> I must find how to page through all the domains, or agree that the report shows only the top domains.

I think the domains could be served in batches. One request calculates the storage needs of the N biggest domain retrieved by the domain facet with an offset param to be able to query the next batch using an infinite scroll on the UI - or someting similar. What do you think?

PerpetualBeta:
Thank you, that is generous. I learned more from the review than from writing the thing.

On batches: I agree with the mechanism, and I think the ordering needs care. A terms facet orders by
document count, and I measured how well document count predicts bytes on my own index of 2110 pages
and 554 domains. The ten largest domains by bytes, with the position each one holds in facet order:

| By bytes | Domain | MB | Pages | Rank by page count |
|---|---|---|---|---|
| 1 | github.com | 20.38 | 434 | 1 |
| 2 | www.bbc.co.uk | 16.37 | 190 | 2 |
| 3 | www.youtube.com | 6.59 | 21 | 13 |
| 4 | en.wikipedia.org | 5.28 | 65 | 7 |
| 5 | news.ycombinator.com | 2.23 | 69 | 6 |
| 6 | edition.cnn.com | 1.74 | 2 | 138 |
| 7 | stackoverflow.com | 1.61 | 53 | 8 |
| 8 | arxiv.org | 1.61 | 1 | 227 |
| 9 | developer.mozilla.org | 1.43 | 102 | 3 |
| 10 | www.figma.com | 1.39 | 7 | 25 |

(Three of the names are stand-ins. All the numbers are real.)

`edition.cnn.com` holds 1.74 MB in two pages and sits 138th by page count. Another domain holds
1.61 MB in one page and sits 227th of 554. A few large pages beat many small ones, and those are the
domains the report exists to find.

Processing domains in facet order, this is how much of the answer each prefix holds:

| Domains processed | Of the true top 20 by size | Share of all bytes |
|---|---|---|
| 10 | 8 | 50% |
| 20 | 11 | 59% |
| 50 | 15 | 67% |
| 100 | 16 | 72% |
| 200 | 17 | 80% |

So a user scrolling in facet order passes 137 rows before meeting their sixth largest domain, and
the server has to process 200 of 554 domains to hold 17 of the true top 20.

There is a smaller point too: `NewFacetRequest` takes a size but no offset. Batch two means asking
for 2N terms and discarding N, and each request repeats the docvalue pass over the index, so the
cost per batch grows with the batch number.

My conclusion is that the batching belongs on the calculation rather than on the presentation. Sizes
live in the documents, so ordering by size means computing every size once. What makes that
affordable is going at the blobs instead of the documents: the store is content addressed, so the
`html_key` and `favicon_key` dictionaries are already the list of distinct blobs. Walking those,
stating each file once, and reading the `domain` docvalue of each posting gives sole ownership for
free and accumulates into a table keyed by domain, which is small. On my index that is 2519 blobs,
154 ms, and 0.5 MB of heap for 554 domains. It can yield every N blobs and report progress. After it
has run once, sorting and paging are free.

For browsing without any size calculation, `FieldDictRange` over `domain` is a real seek into the
FST, so it pages by cursor rather than by offset. Pages of 50 domains measured 60 to 70 µs each and
the cost did not grow with position, and one `Size = 0` term query per row gives the exact page count
at about 4 µs. That gives an instant, exactly paged domain list, with sizes filled in from the pass
above.

One warning if you ever reach for dictionary counts. `DictEntry.Count` is unreliable in the current
zapx: `DictionaryIterator.Next` reuses one `PostingsList`, and `read` does not clear the 1-hit state
when it reads a regular postings list, so `Count` returns 1 for every multi-document term after the
first single-document term in the same scan. On my index a full scan of `domain` reports 2 for
`github.com`, which has 434 pages, and 143 of 555 counts are wrong. Isolated range iterators are
correct, and facets are not affected. hister does not use `FieldDict` anywhere today, so nothing is
broken now. I have a minimal reproduction and a two-line fix and will send them to blevesearch.

I will write this up properly on #534 with the numbers and the trade-offs, including what to leave
out of a first version. Does the shape above sound reasonable to you before I do?

asciimoo:
Interesting thoughts with so many details. I'll need some time to react to it properly.

In the mean time I have an idea about a completely different approach:

Handle the size tracking when modifying documents. Each document insert/update/delete action would update the storage size of domains tracked in the SQL database. So all the size requirements would be available all the time and we could simply order it by size instead of relying on domain page counts.
This approach has downsides as well of course - like the required extra computation for each write operation.

Not sure if this is the right way, just wanted to share.

PerpetualBeta:
Intriguing. I'll do some tests and measurements on that and I'll come back to you.

PerpetualBeta:
I have done some testing and I have to say I like your suggestion better than what I proposed. It
removes the ordering problem instead of working around it: if the sizes are already in SQL then the
report is one `ORDER BY`, and paging and sorting stop being index problems. I measured the read on a
table of 554 domains and it is 0.6 ms for a sorted page of 50.

On the write cost, it depends on one thing that is easy to miss. A document write reaches SQL today
only when embeddings are on, to enqueue the job, so for everyone else this adds a transaction where
there was none. The SQLite DSN also sets no pragmas, so the default is a rollback journal with
`synchronous=FULL`. I measured the same upsert three ways, on an SSD:

| | Per write |
|---|---|
| Default pragmas, one commit per write | 520 µs |
| WAL and `synchronous=NORMAL`, one commit per write | 14 µs |
| Same statement inside an existing transaction | 3 µs |

So the extra cost is either negligible or dominant, depending on where the statement sits. On a
100 000 document import that is 52 seconds of added fsync as things stand, or about a second with
WAL, or a third of a second if the update joins the batch the importer already has open. The idea
looks fine to me on cost, but it wants WAL or a batch-scoped transaction rather than a commit per
document.

One suggestion on where to hook it. If the delta comes from the document you have to reason about
shared blobs, because a second page with identical HTML costs nothing on disk and a domain that
loses one page may release nothing. Both of those are already known one level down:
`writeFileLocked` returns early when the file exists, so it knows whether it created anything, and
`deleteIfOrphaned` already calls `countKeyRefs`, so it knows whether it removed anything. If the
counter moves only when a file is actually created or actually removed, then the sum of the column
equals what is on disk by construction, and there is no refcount bookkeeping in the write path at
all. It needs the domain passed down into the store, which is a small signature change.

That rule has one oddity worth naming: the bytes stay charged to whichever domain caused the file to
exist, so a domain can hold a charge for a blob that only another domain still references. The total
stays exact, the attribution is occasionally generous to the second domain.

Two things the SQL table needs regardless, and they are the same thing: every existing install needs
a first computation, and any counter drifts after a crash or a reindex, so it needs something to
reconcile against. That is what the blob dictionary pass is good for. So I do not think these are
competing designs. Yours is the read path, mine is the backfill and the repair.

I will write both up on #534 rather than keep this going here. Thanks for your time.

PR #709, opened 2026-09-01
adds crawler queue and rate limiting

nburns:
First of two PRs split out from #645; see #710 for the follow-up.

cc @JonnieCache @asciimoo per feedback.


nburns (commented):
Self-review — 3 high, 4 medium, 1 low. Highs are real bugs.

## High

**1. Cancelled fetches mark rows `failed`, not `pending`** — `server/crawler/queue.go:214`
When `fetchCtx` grace-period-expires mid-fetch, `fetchOne` returns `completion{err: ctx.Err()}` and `sqliteQueue.Complete` marks the row `failed`. On resume, `ResetInProgressCrawlURLs` won't touch it (already `failed`), so the URL is silently dropped. Fix: add an `interrupted bool` to `completion`; map it to a `pending` reset in `Complete`.

**2. `q.Complete(ctx, ...)` uses the wrong context** — `server/crawler/crawler.go:210`
Uses the outer caller's `ctx` rather than `context.Background()`. Currently harmless because `Complete` ignores the context, but if that ever changes, a cancelled `ctx` deadlocks all workers (inflight counter never decrements → `Pop` blocks forever on condvar). DB writes here must succeed even after the crawl is cancelled — pass `context.Background()`.

**3. `TryReservePage` load-then-add races and overshoots the budget** — `server/crawler/coordinator.go:332-355`
`Load` then `Add` isn't atomic; N concurrent workers can each see `cur < limit` and all `Add`, overshooting by up to N-1. Fix: `Add` first, then check; on overshoot `Add(-1)` and return false. Or CAS-loop.

## Medium

**4. Fetcher `sem` in chromedp/bidi is dead gating**
Sized to `Rate.GlobalConcurrency`, same as the worker pool. Every worker already holds a `coord.inflightCh` slot before calling `fetchPage`, so the fetcher sem never blocks. Either delete it or hoist the acquire above `coord.Wait` if the intent was to cap total browser tabs across hosts distinctly from worker count.

**5. `sqliteQueue.closed` is set permanently by a side-goroutine watcher** — `queue.go:190-195`
A transient cancellation poisons the queue for all future `Pop` calls. Check `ctx.Err()` directly instead of maintaining shared state mutated across calls.

**6. `Coordinator.Cooldown` TOCTOU on `coolUntil`**
`getHostEntry` acquires+releases the lock, then re-acquires to write `coolUntil`. Between the two, LRU eviction can orphan the entry and the cooldown is lost. Fix: single lock scope via `getHostEntryLocked`.

**7. `getHostEntry` / `getHostEntryLocked` duplicate 35 lines of creation logic**
Correctness is OK today; maintenance hazard for anyone adding per-host state later. Locked variant should do the work; unlocked variant should just lock and delegate.

## Low

**8. `memoryQueue.closed` has the same fragility as sqlite's** — currently safe because all workers share `crawlCtx`; trap for anyone passing per-worker contexts.

## Not an issue

- `PRAGMA journal_mode=WAL` is idempotent, safe on existing installs.

asciimoo:
It looks good overall, thanks. I have a few notes:

- Validation now happens while discovering links, before deduplication, and never checks the starting URL. MaxLinks: 1 can therefore fetch an additional page, while duplicate links consume the limit. Existing callers that read only one document can leave workers blocked sending the extra result.
- Two workers can select the same pending row, then both update it by ID and fetch it. Wrapping these statements in a transaction does not provide exclusive ownership. This needs locking or a conditional update with ownership verification.
- Budget exhaustion cancels `crawlCtx`, but completion checks the original `ctx`. A job with pending URLs consequently becomes completed, and the CLI refuses to resume completed jobs.
- Waiting workers can bypass Retry-After. They check cooldown before waiting for rate tokens and the concurrency slot. If another request subsequently installs a cooldown, those workers proceed without checking again. Slow requests also let workers accumulate tokens and then fetch in rapid succession.
- The breaker returns an error that becomes a permanent URL failure. During an outage, the crawler can drain the remaining host queue into failed status instead of deferring requests until recovery. Another possible solution is a `--retry-failed` switch or dedicated command for crawl jobs. Not sure if solving this issue is in the scope of this PR, but wanted to mention.

nburns:
Thanks for the review, all good points. I've fixed those and found a few more.

nburns:
also found a couple of issues with the suite that I fixed in:
- https://github.com/asciimoo/hister/pull/740
- https://github.com/asciimoo/hister/pull/741

nburns:
Any other feedback or splitting requests?

asciimoo:
Sorry, it's hard and time consuming to review & test a PR this size.
I have the following observations:

- Queue failures can still report success. Errors from `Pop` and `Complete` are logged without propagating them. The driver can mark a job completed despite failing to save results or discovered links. An injected write failure returned success.
- Breaker deferral defeats the retry limit. Requeuing a URL resets its attempt counter. With `max_attempts: 3`, my test fetched one URL nine times before cancellation. During a sustained outage this can continue indefinitely. I think attempt counts need to survive deferral.
- Reaching `MaxLinks` cancels already reserved work. With four workers and `MaxLinks: 2`, only the seed was fetched: another worker exhausted the reservations and cancelled the worker waiting for its rate token. Budget exhaustion should stop new additions while allowing reserved work to finish.
- `--delay` no longer takes effect.
- `max_duration` is not enforced during waits

PR #712, opened 2026-09-01
Fix similarity threshold not adjustable when no results found

kesonglab:
When semantic search returns no results (for example when the similarity threshold is too high), the similarity and weight sliders were hidden because the whole toolbar only shows when there are results. Now the sliders are also shown in the "no results" state, so you can lower the threshold and get results back.

### DIFF
--- a/webui/app/src/routes/+page.svelte
+++ b/webui/app/src/routes/+page.svelte
@@ -566,6 +566,16 @@
   let facetsLoading = $state(false);
   let filtersDropdownOpen = $state(false);
   let actionsDropdownOpen = $state(false);
+  // Auto-open the Actions dropdown when a search returns no results so the
+  // similarity threshold slider (and other actions) are reachable to recover
+  // or refine the query. We latch the user's manual close so we don't fight
+  // them, and reset the latch on each fresh search (see renderResults).
+  let userClosedActions = $state(false);
+  function handleActionsOpenChange(open: boolean) {
+    if (!open) {
+      userClosedActions = true;
+    }
+  }
   let sortDropdownOpen = $state(false);
   // Maps facet name (e.g. "domains", "languages") to the requested top-N size.
   let facetSizes = $state(new Map<string, number>());
@@ -850,6 +860,15 @@
       autocomplete = (query && res.query_suggestion) || '';
       highlightIdx = 0;
       if (!isComposing) resultsShown = true;
+      userClosedActions = false; // reset latch on each fresh search
+      // Guard against out-of-order WebSocket responses: only auto-open if this
+      // response is for the query that's currently in the input. Stale responses
+      // (for a query the user has already moved past) must not open Actions.
+      // SearchResults.query is echoed back as { text }, so compare against
+      // query?.text (a fresh-search response always sets lastResults to itself).
+      if (query && !hasResults && !userClosedActions && lastResults?.query?.text === query) {
+        actionsDropdownOpen = true;
+      }
     }
     hasMore = !!res.page_key;
     pageKey = res.page_key ?? '';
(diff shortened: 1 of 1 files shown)


asciimoo:
Thanks for the contribution. LGTM.
Once lowering the threshold produces a result, the controls immediately move back into the Actions menu. That can interrupt an active slider drag. I'm not sure what would be a good way to handle it. Any ideas?

kesonglab:
Good catch — thanks for the review. I've addressed the mid-drag teardown issue in the new commit (f9aa2b7).

**The fix:** debounce the re-search triggered by the similarity threshold. Previously each slider step fired `sendQuery` immediately, and as soon as the threshold dropped low enough to return results, the empty-state view (and its sliders) were unmounted mid-drag. Now the re-search only fires 300ms after the slider has been left alone (via a `setTimeout` with `clearTimeout` re-armed on every change), so:

- while dragging, the UI stays in the empty-results state and the slider keeps focus,
- after you release, the search runs once and the results view (with the sliders back in the Actions menu) appears cleanly.

The timer is also cleared on component destroy via `onDestroy` so nothing leaks on navigation.

Full diff: `webui/app/src/routes/+page.svelte` — only the threshold `$effect` changed; the empty-state slider block from the first commit is untouched.

kesonglab:
Following up with a review pass on this change — two findings landed as new commits (`2df5361`):

**1. The weight slider was inert in the empty state.** `semanticWeight` only re-ranks already-fetched results client-side and is never sent to the server, so with zero results dragging it did nothing and gave no feedback — sitting right next to the one control that actually helps. The empty state now shows only the threshold slider.

**2. Stale-query debounce.** The debounce effect now reads `query`/`connected` synchronously as tracked dependencies, and a query change cancels a pending timer (the new query's own search fires immediately elsewhere). Only an actual threshold change schedules a delayed re-search.

**Also:** both slider blocks (empty state and the Actions menu) now share a single `{#snippet semanticSliders(showWeight)}` so the duplicated markup can't drift — the menu renders it with `showWeight=true`, the empty state with `false`.

Verified with `svelte-check` (0 errors / 0 warnings on `+page.svelte`) and a clean production build. Still one file touched.

asciimoo:
Wouldn't be better to keep the slider as long as the user has focus or active pointer on the slider instead of using a 300ms debounce?

kesonglab:
Good point, the debounce isn't the right approach. A timer can't tell when a drag actually ends, so it's fragile. Keeping the slider while the pointer or focus is still on it and searching only on release is simpler and avoids that. I'll rework it.

kesonglab:
Swapped the timer for a release/blur trigger. Searches once, and the slider stays put while you're dragging it.

asciimoo:
As I see, the keyboard changes can stop triggering searches after pointer use.
A dirty flag could be set from `oninput`, then search on pointer release or blur only when dirty. This also avoids searching when someone merely focuses and leaves the slider unchanged.

Overall this implementation is becoming quite complex and it has multiple tricky nuances. Wouldn't be better to change direction and keep the sliders only in the "actions" dropdown? It would solve all these edge-cases that can lead to hard to debug bugs. Perhaps we can go one step further and automatically expand the dropdown in case of having no results.

What do you think?

kesonglab:
Going with your suggestion, way simpler and no more timing edge cases. Force-pushed the branch down to a single clean commit on top of master.

The change is just an `$effect` that opens the Actions dropdown the instant we land in the empty-results state. There is a `wasEmpty` latch guarding it so we do not reopen it if the user closes it manually. The threshold `<input>` inside the Actions menu is untouched. No empty-state slider block, no snippet, no drag/pointer tracking, no debounce. Mid-drag teardown goes away too, since the dropdown is not unmounted when results come back.

commit `c32cd76`.

`vite build` is clean, no new tsc errors in `+page.svelte`. The pre-existing `fallow` lint failures and `webui/components` tsc errors on master are not related to this change.

asciimoo:
Have you tested these changes?
Are you using AI for the development?

The effect sets `actionsDropdownOpen = true`, but the menu is nested inside `{#if hasResults}`. When the empty state is active, that entire menu is unmounted, so changing its open state has no visible effect.

Also, `noResults` only checks `lastResults.documents.length`. The UI also counts history and semantic only matches through `mergedResults`, so this can open Actions even while visible results exist.

Also, `wasEmpty` is shared across searches. Since `lastResults` is retained while a new nonempty query is loading, two consecutive empty searches do not create a `false` to `true` transition. If the user closes Actions for the first query, it will not reopen for the second.

kesonglab:
Yeah, all three were right. Sorry about that, `ebe51fc` should not have gone up. The dropdown was inside `{#if hasResults}` the whole time, so setting `actionsDropdownOpen = true` in the empty state was literally a no-op. Both the predicate and the latch were off too. All of it was obvious from the template, I just did not read it carefully enough.

I had Copilot help me draft v1, and that is almost certainly how this slipped. The suggestion looked plausible, I went with it, never opened the browser. Should have caught that before pushing.

For `43277fd`:

- Moved the `<DropdownMenu.Root>` block out of the toolbar and out of `{#if hasResults}` entirely, so it is always mounted at the top of the results column
- Auto-open moved into the fresh-search branch of `renderResults`, keyed on `!hasResults` (same `$derived` the UI uses, so history and semantic-only hits count correctly)
- `userClosedActions` resets on each fresh search; an `onOpenChange` handler latches it on close, so a manual close sticks until the next search
- Stale-response guard: `lastResults?.query?.text === query` so an out-of-order WebSocket response for an older query cannot auto-open Actions

Tested for real this time with Playwright + a mocked WebSocket. Five scenarios, all passing:

```
01 empty-results search auto-opens the Actions dropdown
   - expect(actionsTrigger).toBeVisible()
   - expect(actionsContent).toBeVisible()
   - expect("No results found for").toBeVisible()

02 open Actions content exposes the similarity threshold slider
   - expect(slider).toBeVisible()
   - expect(slider).toHaveAttribute("min", "0")
   - expect(slider).toHaveAttribute("max", "1")

03 user-close is respected until the next fresh search
   - close via Escape, content stays hidden
   - submit a new empty search, content re-opens

04 non-empty search keeps the Actions dropdown closed
   - example result visible, actions content count == 0

05 stale WebSocket response for an older query does not auto-open Actions
   - inject an older query response after the user has moved on
   - Actions stays closed

asciimoo:
Actions block is now unconditional within the search screen. Browser navigation can produce an empty query while resultsShown remains true, because `handlePopState` clears `lastResults` but not `resultsShown`. In that state, Actions is visible with no query. Selecting Delete all then converts the empty query to `*`, potentially deleting the entire index, while the confirmation displays zero results and a blank query.

Also, the stale response guard compares lastResults.query.text with the raw input. The server replaces aliases before returning the query, so an alias such as docs resolving to type:file makes the comparison fail. An empty search made through an alias will therefore not open Actions.

There is also a smaller cleanup issue: `userClosedActions` is reset to false immediately before its only decision point, so the latch does not affect behavior.

I would keep toolbar mounted when `query && lastResults`, conditionally show result specific actions, and correlate responses using an original query value or request identifier.

PR #718, opened 2026-09-02
[fix] make the Nix vendorHash updater detect stale hashes reliably

samhza:
The Go dependencies are stored in a fixed-output derivation, so before this change cached output could get reused.  This meant that the vendorHash would not be updated and that the script would also [fail](https://github.com/asciimoo/hister/actions/runs/32754604974/job/97519107285) to build hister because it was building against outdated dependencies.

The update script now uses a fake hash to force a re-vendor.

Add flake.lock as a trigger so the hash is updated whenever the Go toolchain is updated.

Drop the node/NPM references from the workflows and update scripts because Nix vendors those hashes directly from node's lockfiles since bfb6ecb.

AI disclosure: Claude helped rewrite the script. I fully understand the changes made.

### DIFF
--- a/.forgejo/workflows/update-nix-vendor-hash.yml
+++ b/.forgejo/workflows/update-nix-vendor-hash.yml
@@ -37,10 +37,10 @@ jobs:
         with:
           token: ${{ secrets.FORGEJO_TOKEN }}
           branch: update-nix-hashes
-          commit-message: '[chore] update Nix hashes for Go/NPM changes'
-          title: 'Update Nix hashes'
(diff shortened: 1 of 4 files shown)


4evy:
This fixes the immediate bug, but imo the custom script is the wrong abstraction. `nix-update` already handles hash invalidation, package edits, and build verification, while `passthru.updateScript` keeps that logic with the package instead of burying it in CI glue

I opened a PR to replace the script entirely: https://github.com/asciimoo/hister/pull/721

PR #721, opened 2026-09-02
[chore] use `nix-update` for dependency hashes

4evy:
`nix-update` handles fixed-output hash mismatches, edits `package.nix`, and builds the result. `passthru.updateScript` makes that complete recipe package metadata, so local runs, nixpkgs-update, and CI all use the same entry point

This removes a bespoke script that scraped Nix error messages, replaced matching text blindly, and stopped after two mismatches


asciimoo (approved):
Thanks!

Issue #172, opened 2026-03-06
no analyzer with name or type 'cs' registered


shaman007:
Nightly (?) build. Run hister reindex.

 | WARN  | indexer/indexer.go:248 > Failed to create language indexer error="no analyzer with name or type 'cs' registered" Name=index_cs.db

I don't know is it important or not.

Bthw, please make nighly builds really nighly.

PR #724, opened 2026-09-04
[fix] do not create indexes for languages that Bleve cannot analyze

samhza:
Bleve has no analyzers for bg, ca, cs, el, eu, ga, hy and id, only stopword lists.  Creating an index for these languages failed with "no analyzer with name or type 'ca' registered" (see for example #172) and logged a warning for every document.  These languages were already sent to the default indexer before this change, they now go there without emitting warnings.

Remove the bleve imports for these languages because they are unused.

### DIFF
--- a/server/indexer/indexer.go
+++ b/server/indexer/indexer.go
@@ -1416,7 +1416,7 @@ func (i *Indexer) getOrCreate(lang string) bleve.Index {
 }
 
 func indexNameForLanguage(lang string) string {
-	if lang == document.UnknownLanguage || lang == "" {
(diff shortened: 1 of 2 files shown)


asciimoo:
Thanks for the contribution.

The new `languageHasAnalyzer` is a blacklist. Unknown values such as `zz` or `default` return true, so Hister still attempts to create an index with a nonexistent analyzer.

I would use:
```golang
if lang == document.UnknownLanguage || lang == "" || !registeredLanguageAnalyzer(lang) {
    return defaultIndexerName
}
```

asciimoo (approved):
Great, thanks!

PR #726, opened 2026-09-05
Add support for .eml files

HiemSword:
Add support for indexing and previewing .eml files. 
I used emersion/go-message for the eml parsing. Email metadata such as sender and cc are saved in the metadata of the document
Tested with emails from proton and the linux mailing list

Note:
In the current state some emails that use a lot of styling are not displayed properly, will create a proper preview function in the extractor later.

Disclaimer: I used AI to assist with this PR, but I wrote and reviewed all the code myself.


HiemSword:
Also I added tests for the indexer similar to the ones for markdown, but when I tried it with the live cases it dosen't seems to work? The crawler is blocked in almost every site and can't fetch. Idk if this is just on my machine or a known issue

PR #740, opened 2026-09-10
[fix] store database timestamps in UTC

nburns:
SQLite has no date type: it stores a time as text and compares it as text, and the driver stores "whatever timezone they come with". `gorm` generated `CreatedAt`/`UpdatedAt` from `time.Now()`, so a server west of UTC wrote `2026-09-09 20:21:05.855826-07:00` while queries bind their ranges in UTC (`2026-09-10 00:00:00+00:00`). Compared as text, `2026-09-09 ...` sorts before `2026-09-10 ...`, so the row falls outside the UTC day it belongs to.

Visible effect: the history timeline reports an item for today, and drilling into that same day returns nothing. It affects any server whose local date differs from the UTC date, and it is invisible on a UTC machine, so CI never caught it. `TestPublicModeEnablesHistoryForAuthenticatedCallers` fails today under `TZ=America/Los_Angeles` and passes under `TZ=UTC`.

## Changes

- Generate gorm timestamps in UTC through `gorm.Config.NowFunc`.
- Convert caller-supplied times to UTC at the model boundary rather than trusting each call site: embedding retry deadlines, web session expiry, and the history pagination cursor.
- Add `TestHistoryDateRangeIgnoresServerTimezone`, which pins its own local zone eight hours behind UTC so it fails regardless of the machine running it.

## Notes

Rows written before this change keep their original offsets, so their text order stays wrong until they are rewritten. Fixing those needs a migration that rewrites every timestamp column, which is not included here.

Verified with the full suite under `TZ=UTC`, `America/Los_Angeles`, `Asia/Tokyo` and `Australia/Sydney`.

### DIFF
--- a/server/model/embedding.go
+++ b/server/model/embedding.go
@@ -52,7 +52,7 @@ func EnqueueEmbeddingJob(docID string) error {
 	if err != nil {
 		return err
 	}
-	now := time.Now()
+	now := time.Now().UTC()
 	job := EmbeddingJob{
 		DocID:       docID,
 		Status:      EmbeddingJobPending,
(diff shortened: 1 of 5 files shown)


asciimoo:
Great, thanks!

PR #741, opened 2026-09-10
[fix] watch symbolically linked directory roots

nburns:
A directory root reached through a symbolic link (`~/notes -> /mnt/data/notes`) was watched by its link path, which breaks in two ways.

**Events are reported under the target.** fsnotify's kqueue backend follows the link and watches the target, so macOS and the BSDs report `/mnt/data/notes/file.md` while the configured root — and every filter derived from it — speaks in `~/notes/...` paths. Nothing matched, so no file under such a root was ever reported. inotify reports events under the name the watch was registered with, so Linux (and CI) never saw this.

**Nested directories were never watched.** `filepath.WalkDir` does not descend a symbolic link on any platform, so the scan that registers watches stopped at the root. Only direct children of the root produced events, and only on Linux.

`TestWatcherSymlinkDirectoryRoot` fails on macOS today; the new nested-directory test fails on Linux too.

## Changes

- Resolve each configured root once and keep it alongside the path the user configured.
- Register watches and walk the tree at the resolved location, translating event paths back into configured paths before filtering, scheduling, or reporting — so callbacks still return the path that was configured, not the link target.
- Add `TestWatcherSymlinkDirectoryRootSubdirectory` for the nested case.

Non-symlinked roots resolve to themselves, so nothing changes for them.

### DIFF
--- a/files/watcher.go
+++ b/files/watcher.go
@@ -24,8 +24,19 @@ const debounceTime = 200 * time.Millisecond
 type Watcher struct {
 	watcher *fsnotify.Watcher
 	dirs    []*config.Directory
+	links   []symlinkRoot
 	paths   map[string]bool
 	strict  bool
+}
+
+// symlinkRoot records a configured root that lives somewhere else on disk.
+// Watches and directory walks have to use the resolved location - kqueue
+// follows the link and reports events under the target, and WalkDir does not
+// descend a symbolic link at all - while filters and callbacks keep using the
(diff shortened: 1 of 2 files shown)


asciimoo:
Nice, thanks. Please fix the linting and it is ready to merge.

asciimoo:
Sorry, seems like the lint issues are unrelated. I'm fixing them.

PR #764, opened 2026-09-19
Add browser extension links to the hamburger menu

bwhaley:
The README asks users to install the Firefox or Chrome extension, but the web UI gives no way to reach either store page. This adds both links to the hamburger menu so a user who opens Hister for the first time can find them without leaving the app.

## Changes
- Add "Firefox extension" and "Chrome extension" entries to the secondary menu in `webui/app/src/lib/components/SiteHeader.svelte`, after GitHub.
- Both use the existing external-link handling: new tab, `rel="noopener"`, external-link icon.

## Verification
- `npm run build -w @hister/app` succeeds.


asciimoo:
It would be better to display only the link that matches the browser. Perhaps, it would be even better to hide it completely if the extension is already installed.

bwhaley:
Displaying only the link that matches the browser is easily done, but hiding it completely if the extension is already installed is more of a challenge and has tradeoffs. The options I can see:

1. The extension exposes a file via `web_accessible_resources` in its manifest, and hister can attempt to fetch that specific URL. Limitation: Only compatible with Chrome.

2. `chrome.runtime.sendMessage` - also only compatible with Chrome

3. The extension injects a market in to the DOM, and hister detects the marker. Limitation: Any web site could detect that the user runs hister.

A compromise could be option (3), but only insert the market on the default server URL. Users running non-default URL would always see the link in the menu.

Given these constraints, do you have a preference? 

Fwiw, my primary motivation for this change is that I had already loaded Hister on first run and expected to see a link to the chrome extension within the web app, but I had to visit the installation instructions instead. I also run multiple Chrome profiles, so I had to install the extension a few times. Not a biggest, so it's fine if you just don't want to do it.

bwhaley:
Added the change to only show the link that matches the browser, in case you want to go that direction.

asciimoo:
The extension detection is out of scope of this PR but it is a good future enhancement.

> The extension injects a market in to the DOM, and hister detects the marker. Limitation: Any web site could detect that the user runs hister.

The extension knows the server URL (otherwise it could not submit the documents), so the extension can detect if the opened page is the configured Hister service and inject the marker only in these cases.

----

The PR looks good, thank you for your contribution

PR #765, opened 2026-09-19
feat: support reverse proxy authentication via configurable header

### LINKED ISSUE #761: Support Forward/Proxy Auth
It would be great to be able to have authentication occur in an external service/proxy. This ensures that not a single request can make it to Hister before it is properly authenticated. This allows easy and consistent authentication as well as ensures that even if there is a vulnerability in Hister it isn't exploitable by anonymous users.

The simplest form would be to read the `Remote-User` header (or some configurable alternative) for the username and handle that request authenticated by that user. This header would be set by the proxy (and never allowed to be set from the internet). This obviously needs to be an opt-in configuration option to avoid being insecure by default.

akshayvibe:
Closes #761

### Summary
Adds optional authentication via a trusted reverse proxy using a configurable HTTP header such as `Remote-User.`
**server:**
 ` proxy_auth_header: "Remote-User"`

When configured:
- user_handling is enabled automatically.
- Hister uses the header value as the authenticated username.
- Existing users are used directly.
- New users are automatically created on their first request.
- Missing or empty headers are treated as unauthenticated.

When unset, Hister behaves exactly as before.
The feature is opt-in and assumes Hister is behind a trusted authentication proxy. The proxy must strip the configured header from incoming requests so clients cannot spoof the authenticated user.

### Changes
- Add `ProxyAuthHeader` to server configuration.
- Add proxy authentication to `populateUserContext()`.
- Automatically create users authenticated by the proxy.
- Add tests for existing/new users, missing headers, disabled config, and custom headers.


asciimoo:
Nice, thanks. Few notes:

 - New users receive a valid empty password. `CreateUser(username, "", false)` hashes the empty string. Those credentials become usable for normal access if proxy auth is later disabled while user handling remains enabled. Proxy accounts should have password authentication disabled, as OAuth accounts already do.
 - The tests do not establish that authentication works. All six pass even after removing the entire new authentication block.
 - Please preserve API token authentication when proxy auth is enabled. Currently `populateUserContext()` returns immediately when the configured proxy header is missing or empty, bypassing both Bearer and `X-Access-Token` authentication. This breaks token access for CLI, extension, and MCP clients.

Issue #737, opened 2026-09-09
[Feature Request] Add optional recursive "Crawl this site" mode to the browser extension (with full control over limits, including MAX)


archerhush:
# The missing piece: recursive crawling in the browser extension

You're overlooking the fundamental point of a (public) search engine: to cover the open web, it **must** recursively follow links from page to page. Without that, it can never discover more than a tiny, static set of URLs. Hister, by design, is a private search engine over your own visited pages and files, with optional imports and targeted crawls. Adding a true recursive indexer would make it far more meaningful, since it could start indexing pages that other search engines might never reach.

The most important capability of any web search engine is a **recursive crawl** (recursive traversal) of the web graph.

## Why recursion is central

Search engines don't have a master list of all URLs. Instead, they start from a small set of "seed" pages and then:

1. Fetch a page  
2. Parse it to extract all outgoing links  
3. Add those links to a queue  
4. Repeat the process for each new link  

This loop—fetch → parse → extract links → enqueue → repeat—is inherently recursive: every newly discovered page can lead to more pages, which can lead to even more, and so on.

Without recursive crawling:

- The engine would only know about manually submitted URLs or a tiny static list.  
- It could not discover the vast "long tail" of the open web, including new or updated content.  
- Index coverage would be extremely poor, making search results incomplete and much less useful.

In other words, recursion is what lets a search engine scale from a few known pages to billions, by treating the web as a giant directed graph and systematically exploring it.

Hister currently indexes *your* visited pages and local files (with optional browser-history import and targeted crawls), not the open web via continuous recursive link following. That makes it a privacy-focused personal search tool rather than a traditional public web search engine. Adding a recursive crawler would let it go beyond your own history and potentially surface pages that mainstream engines haven't indexed yet.

> **Note:** Recursive crawling *is* already possible via the CLI (`hister index --recursive ...`), but **not** from the browser extension. This feature request is specifically about bringing that capability into the addon, where it can reuse the browser's session, cookies, headers, and rendering.

---

# Feature request: Add optional recursive "Crawl this site" mode to the browser extension (with full control over limits, including MAX), which can be enabled and disabled with Hotkeys too

## Summary

Add an **opt-in, user-triggered recursive crawl** feature to the Hister browser extension that lets users turn the current page into a seed for a controlled crawl of the site/domain. Crucially, this must be **fully customizable**, including:

- Maximum number of links to follow (**must allow `MAX` / unlimited**)  
[...]

PR #767, opened 2026-09-20
Add site crawling from the browser extension

rasulkireev:
The extension can index the current page, but cannot start a crawl of the site. This adds **Crawl this site** to the popup, with a page limit, progress, and a Stop button. Crawling continues after the popup closes; reopening it retrieves the current status.

Related to #737. This PR implements a bounded, server-side crawl of public pages; browser-session reuse, JavaScript rendering, and the issue's broader configuration options are outside its scope.

## Screenshot

Hister extension popup with matching Index this page now and Crawl this site buttons in the same section

Built extension popup shown during the Playwright check, using mocked Chrome APIs. The crawl button sits directly beneath the page-indexing button and shares its styling.

## Approach

- Reuse the existing crawler's breadth-first traversal, HTTP fetching, link deduplication, robots.txt checks, and delay. The extension calls authenticated start/status/stop endpoints; it does not implement another crawler or invoke the CLI as a subprocess.
- Keep the existing CLI crawler behavior unchanged: the new error reporting and HTTP connection cleanup are isolated to the site crawler. Shared extraction and indexing changes are additive helpers used by this feature.
- Start from the current site's home page and follow links on the exact hostname. Use one in-memory crawl per server to avoid adding a database migration or worker system. Indexed documents persist, while job status is discarded on restart or replacement.
- Apply the current user's skip rules and document ownership. Fetch already indexed pages to discover links, but preserve their existing documents. Redirect targets must also satisfy hostname, skip-rule, and robots.txt checks.
- Restrict server-side fetching to public HTTP(S) addresses on standard ports. Resolve and dial checked IP addresses to prevent DNS rebinding into private networks. These checks also cover redirects and robots.txt requests. Do not forward configured cookies, headers, or proxies.
- Use the existing readability/basic HTML extraction with sensitive-content checks. Avoid network-based enrichers that could fetch additional URLs outside the crawl restrictions.

The defaults are 100 page visits, with a maximum of 1,000, a one-second delay, and a 30-minute deadline. Discovery is capped at 10,000 URLs and responses at 5 MiB. Sitemap import, JavaScript rendering, automatic retries, subdomain traversal, and restart/resume are outside this first implementation. This follows discoverable links rather than guaranteeing every page on a domain is indexed. Both the server and extension need this change; the popup explains when the server does not support it.

## Testing

Completed locally on macOS ARM64:

- App and extension production builds, plus `go build`.
- `npm test -w @hister/ext`: all 22 existing extension tests passed.
- `golangci-lint run --fix ./...`: zero issues; Prettier checks passed for the changed Svelte files.
- New backend tests passed with the Go race detector: `go test -race ./server/crawler ./server -run 'Test(PublicSite|Site)' -count=1`. Coverage includes request authentication/CSRF, ownership, stop/shutdown, concurrent-start rejection, duplicate preservation, page limits, robots/skip rules, redirect restrictions, private/reserved addresses, and response size limits.
- `go test ./...` has one failure: `TestResolveHistoryImportsRejectsUnknownBrowser`. The same test was reproduced failing on unmodified upstream commit `0cd1ef5`. The full suite passes with that test excluded using `-skip '^TestResolveHistoryImportsRejectsUnknownBrowser$'`.
- After auditing scope, `go test -race ./server/crawler ./server -count=1` and `go test ./cmd -skip '^TestResolveHistoryImportsRejectsUnknownBrowser$' -count=1` passed. A regression test checks that error reporting and connection cleanup remain specific to site crawls. Lint was rerun with zero issues.
- A real crawl through the new API on a local server indexed two pages from hister.org with a limit of two.
- Playwright exercised the built popup with mocked Chrome APIs and crawl responses: start payload/page limit, progress, reopening during a crawl, stopping, and the unsupported-server message. This was not a live Chrome extension installation or store-release test.

- After the popup layout adjustment, the extension build, all 22 extension tests, and Prettier checks passed again. Playwright verified matching button dimensions, typography, colors, and borders in the same section, plus expand/start/progress/stop with mocked API responses.

## AI disclosure

OpenAI Codex authored the implementation, tests, and this PR description at the submitter's request.


rasulkireev:
Human note:
- tested this locally (extension) with and custom deployment on `hister.lvtd.dev`
- tried to leverege the existing code that supports site crawling as much as possible.

rasulkireev:
+ couple of screenshots from using it locally:
CleanShot 2026-09-20 at 12 22 43@2x
CleanShot 2026-09-20 at 12 22 20@2x
CleanShot 2026-09-20 at 12 22 10@2x

asciimoo:
Currently we don't perform any kind of untrusted requests from the server. The proposed solution would break it.
What about generating only the `hister index` command required to run a crawl job from the extension with all the necessary arguments?
It would be an extra step for the user, but we'd keep the server safer and more reliable.

